PR #5, opened 2023-11-10
feat: port cacerts rule

thesayyn:
cc @alexeagle

### DIFF
--- a/WORKSPACE.bazel
+++ b/WORKSPACE.bazel
@@ -1,1 +1,12 @@
 # Marker that this is the root of a Bazel workspace.
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "ca-certificates",
+    build_file_content = 'exports_files(["data.tar.xz"])',
+    canonical_id = "test",
+    sha256 = "b2d488ad4d8d8adb3ba319fc9cb2cf9909fc42cb82ad239a26c570a2e749c389",
+    type = ".deb",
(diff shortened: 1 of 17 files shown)


distroless/private/cacerts.bzl
```diff
@@ -0,0 +1,80 @@
...
+            cacerts.path,
+            copyright.path,
+        ],
+    )
+
+    output = ctx.actions.declare_file(ctx.attr.name + ".tar.gz")
+    mtree = tar_lib.create_mtree(ctx)
+    mtree.add_file_with_parents("/etc/ssl/certs/ca-certificates.crt", cacerts)
+    mtree.add_file_with_parents("/usr/share/doc/ca-certificates/copyright", copyright)
+    mtree.build(output = output, mnemonic = "CaCertsTarGz", inputs = [cacerts, copyright])
```
thesayyn, line 58: @alexeagle this is awesome. gives me more confidence that bsdtar was a good choice.

WORKSPACE.bazel
```diff
@@ -1 +1,12 @@
 # Marker that this is the root of a Bazel workspace.
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
```
loosebazooka, line 5: Maybe I don't understand bazel, but should this be here? should examples have it's own workspace?

loosebazooka (approved):
looks good

WORKSPACE.bazel
thesayyn: unfortunately yes, module.bazel doesn't have an easy way to use http_archive in module.bazel files.

thesayyn: this not necessary in bazel 7 but we are testing against 6 as well. so we are stuck using this

PR #6, opened 2023-11-11
feat: port locale rule

thesayyn:


distroless/private/locale.bzl
```diff
@@ -0,0 +1,77 @@
+"locale"
+
+load(":tar.bzl", "tar_lib")
+
+_DOC = """Create a ca-certificates.crt bundle from Common CA certificates.
```
loosebazooka, line 5: Doc comment seems off?

WORKSPACE.bazel
```diff
@@ -3,10 +3,22 @@
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 http_archive(
-    name = "ca-certificates",
+    name = "bullseye-ca-certificates",
```
loosebazooka, line 6: I was thinking about this. Is it unusual for us to rename these to indicate they are for test only `test-bullseye-ca-caerts`? Or `example-bullseye-ca-certs`

loosebazooka (approved):

WORKSPACE.bazel
thesayyn: oh yes let's rename them so that they have the example prefix

PR #9, opened 2023-11-17
feat: implement java_keystore

thesayyn:
This is somehow a port of jksutil into Java with parts taken from https://github.com/openjdk/jdk17u-dev/blob/a028120220f6fd28e39fe0f6190eb1f5da6a788d/make/jdk/src/classes/build/tools/generatecacerts/GenerateCacerts.java. 

Using java and java_binary is less tedious than using go and go_binary since java comes bundled with java, therefore, there's no need to take dependency just for this.


thesayyn:
> Does this now become an environment specific build? I'm not sure I understand enough, but keytool is obtained from the build env?

No, we depend on an hermetic jdk toolchain to get the `keytool`, so this will work even if you don't have keytool installed locally.

loosebazooka:
Kk, lgtm after test fix.

thesayyn:
eh, tests are failing because of timezone differences between my local machine and the CI. so interesting to find a new way of non-hermeticity.

alexeagle (approved):

distroless/private/java_keystore.sh
```diff
@@ -0,0 +1,19 @@
...
+
+readonly bsdtar="$1"
+readonly keytool="$2"
+readonly coreutils="$3"
+readonly output="$4"
+shift 4;
+
+tmp=$(mktemp -d)
+while (( $# > 0 )); do
+    $coreutils csplit --quiet --elide-empty-files -f "$tmp/crt" -b "%02d_$#.crt" -k $1 '/-----BEGIN CERTIFICATE-----/' '{*}'
```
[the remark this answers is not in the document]
thesayyn: agreed

thesayyn:
ehhhhhhh, https://bugs.openjdk.org/browse/JDK-8278157

thesayyn:
@loosebazooka PTAL.

examples/java_keystore/BUILD.bazel
```diff
@@ -0,0 +1,37 @@
+load("//distroless:defs.bzl", "java_keystore")
+load("//distroless/tests:asserts.bzl", "assert_jks_listing", "assert_tar_listing")
+
+java_keystore(
+    name = "java_keystore",
+    certificates = [
+        # asserting that we support both bundle x509 certs
+        # and single x509 certs
+        "amazon.crt",
```
loosebazooka, line 9: I know this doesn't matter, but it does feel a bit strange to have the amazon cert in here.

examples/java_keystore/BUILD.bazel
```diff
@@ -0,0 +1,37 @@
...
+    actual = "java_keystore_jks",
+    expected = "expected.jks.output",
+)
+
+assert_tar_listing(
+    name = "test_java_keystore",
+    actual = "java_keystore",
+    expected = """\
+#mtree
+./etc time=1672560000.0 mode=755 gid=0 uid=0 type=dir
```
loosebazooka, line 31: I know you mention what this time was before, but i can't seem to find it and I'm confused again why it's not just 0? What's the significance of this time?

loosebazooka (approved):
lgtm, I think maybe we can make the cert example change here before merge?

examples/java_keystore/BUILD.bazel
thesayyn: i took it from https://github.com/GoogleContainerTools/distroless/blob/be23a86353166b733a7088a35fa357cc79afbe60/cacerts/jksutil/test.crt#L4 and split the amazon ca-cert out of bundle to test both cases.

thesayyn: @loosebazooka wdyt?

loosebazooka: oh lulz, sgtm

examples/java_keystore/BUILD.bazel
loosebazooka: sgtm

PR #16, opened 2024-02-03
feat: implement an apt resolver

thesayyn:


apt/private/index.bzl
```diff
@@ -0,0 +1,120 @@
...
+        build_file_content = """\
+load("@rules_distroless//apt/private:dpkg_statusd.bzl", "dpkg_statusd")
+load("@rules_distroless//distroless:defs.bzl", "flatten")
+
+filegroup(
+    name = "odata",
+    srcs = ["data.tar.xz"]
+)
+
+dpkg_statusd(
```
thesayyn, line 19: This might not be what we want;  newer versions of dpkg doesn't seem to recognize `status.d (./var/lib/dpkg/status.d/*)` files. only recognizes `/var/lib/dpkg/status` file. We might not do this by default.

loosebazooka: `status.d` is a very distroless specific status page: https://github.com/bazelbuild/rules_docker/pull/239, the tl;dr appears to be that it was selected to avoid collisions.

thesayyn:
ping @alexeagle for Bazel API review

apt/private/index.bzl
thesayyn: Then i'll leave this to the users and not generate in the default template. They can always override this template to add more stuff.

.bazelrc
```diff
@@ -21,6 +21,10 @@ build --java_runtime_version=remotejdk_11
 # https://github.com/GoogleContainerTools/rules_distroless/actions/runs/7118944984/job/19382981899?pr=9#step:8:51
 common:linux --sandbox_tmpfs_path=/tmp
 
+
+# Allow external dependencies to be retried. debian snapshot is unreliable and needs retries.
+common --experimental_repository_downloader_retries=200
```
alexeagle, line 26: 200 is a high number! would we actually want a build to succeed after spending the time to make 199 failed downloads?

also, I think this flag probably needs to go into bazel-lib presets, could you make a PR there too?

examples/apt/bullseye.yaml
```diff
@@ -0,0 +1,33 @@
...
+#  To generate the bullseye.lock.json run the following command
+#
+#     bazel run @bullseye//:lock
+#
+# See debian_package_index at WORKSPACE.bazel
+version: 1
+
+sources:
+  - channel: bullseye main
+    url: https://snapshot-cloudflare.debian.org/archive/debian/20240210T223313Z
```
alexeagle, line 14: is it possible to have multiple URLs - one that's more reliable and a second that's more canonical/guaranteed?

examples/apt/test_linux_amd64.yaml
```diff
@@ -0,0 +1,23 @@
...
+commandTests:
+  - name: "echo hello"
+    command: "/bin/bash"
+    args: ["-c", "echo hello world!"]
+    expectedOutput: ["hello world!"]
+  - name: "apt list --installed"
+    command: "apt"
+    args: ["list", "--installed"]
+    expectedOutput:
+      - Listing\.\.\.
```
alexeagle, line 12: nit: is there a yaml syntax that allows this to be less escape-y ?

alexeagle (approved):
high-level bazel code in examples/apt looks okay to me, we could do a sync review over VC if there's anything specific you have in mind.

.bazelrc
loosebazooka: Hah yeah that's pretty high, perhaps we just suggest people use the cloudflare snapshot mirror? and keep this number a bit lower. Distroless will probably be one of the most heavy consumers of this, people building one off will probably only need to download a few packages

MODULE.bazel
```diff
@@ -8,10 +8,18 @@ module(
 
 bazel_dep(name = "bazel_skylib", version = "1.5.0")
 bazel_dep(name = "aspect_bazel_lib", version = "2.5.3")
+bazel_dep(name = "container_structure_test", version = "1.16.0")
+bazel_dep(name = "rules_oci", version = "1.7.4")
 
 bazel_lib_toolchains = use_extension("@aspect_bazel_lib//lib:extensions.bzl", "toolchains")
-bazel_lib_toolchains.tar()
 use_repo(bazel_lib_toolchains, "bsd_tar_toolchains")
+use_repo(bazel_lib_toolchains, "yq_darwin_amd64")
```
loosebazooka, line 16: are these conditional on what OS the user is running on? (is that a thing that is possible with bazel?)

loosebazooka (commented):
looks mostly good.

Want to make sure you handled some edge cases, I remember adding test cases for weird multiline parsing: https://github.com/GoogleContainerTools/distroless/blob/8166efb13d435c5b95bf3ff44b1479bd94403f9b/debian_package_manager/internal/deb/parser_test.go#L48

.bazelrc
thesayyn: Upps, this is leftover from an experiment. Will lower it.

examples/apt/test_linux_amd64.yaml
thesayyn: this bugs me too. container-structure-test doesn't have match this exactly option. everything has to be a regex, so the escapes are side effect of regex

MODULE.bazel
thesayyn: oh yes, these are conditional on where Bazel runs as we run yq in a repository rule, and repository rules are executed on the host, so we have to pick the right one for the host.

examples/apt/bullseye.yaml
thesayyn: can could be a future implementation, we'll be iterating over this a lot before 1.0. i'd rather leave that as a follow-up/future work.

thesayyn:
> Want to make sure you handled some edge cases, I remember adding test cases for weird multiline parsing:

That's done already.

PR #17, opened 2024-02-20
fix: match rules_docker modes

thesayyn:
This fixes the issue where we didn't have matching permissions as rules_docker.


loosebazooka (approved):

loosebazooka:
lgtm

PR #32, opened 2024-04-14
Fix for packages sharing the same dependencies

jacobshirley:
Hi there, I was running into issues where, if multiple packages depend on the same packages, only the first one in the list will receive them. By removing the performance code in the resolver, it appears to be fixed. I've added a test, which verifies the nginx has a module that is required by both 'nginx-full' and 'nginx-core', when only 'nginx-core' goes into the image.

Please feel free to close this PR and do a better fix if there is one.


loosebazooka:
@jacobshirley we require a signed CLA before proceeding

thesayyn (approved):
Can you move the new e2e test into examples/shared_dependencies?

jacobshirley:
Sure, that's done now

examples/shared_dependencies/MODULE.bazel
```diff
@@ -0,0 +1,25 @@
+bazel_dep(name = "rules_distroless", version = "0.0.0", dev_dependency = True)
```
thesayyn, line 1: You don't need a MODULE.bazel here, you can just use existing MODULE at root

examples/shared_dependencies/MODULE.bazel.lock
```diff
@@ -0,0 +1,3543 @@
+{
```
thesayyn, line 1: remove the lock please

examples/shared_dependencies/WORKSPACE.bazel
```diff
@@ -0,0 +1,46 @@
+workspace(name = "shared_dependencies")
```
thesayyn, line 1: Also remove the workspace here.

jacobshirley:
Apologies for the delay on this. Hopefully that's looking ok now

thesayyn (approved):

thesayyn:
Thank you!

Issue #31, opened 2024-04-11
@index//:lock fails when yaml is in workspace root


NotBobTheBuilder:
Hello!

It appears the manifest for `deb_index` must be in a subdirectory for the `:lock` operation to complete successfully. If the manifest is at the workspace root, the `:lock` operation seems to try to write the lockfile into the os root directory `/`.

Here's an example minimum reproduction, and additional example of it working after moving the manifest

WORKSPACE.bazel
```
load("@rules_distroless//apt:index.bzl", "deb_index")

deb_index(
    name = "bullseye",
    # lock = "//:bullseye.lock.json",
    manifest = "//:bullseye.yaml",
)

load("@bullseye//:packages.bzl", "bullseye_packages")
bullseye_packages()
```

MODULE.bazel
```
bazel_dep(name = "rules_distroless", version = "0.3.1")
bazel_dep(name = "aspect_bazel_lib", version = "2.6.0")

bazel_lib_toolchains = use_extension("@aspect_bazel_lib//lib:extensions.bzl", "toolchains")
bazel_lib_toolchains.tar()
use_repo(bazel_lib_toolchains, "coreutils_toolchains")
use_repo(bazel_lib_toolchains, "yq_darwin_amd64")
use_repo(bazel_lib_toolchains, "yq_darwin_arm64")
use_repo(bazel_lib_toolchains, "yq_linux_amd64")
use_repo(bazel_lib_toolchains, "yq_linux_arm64")
use_repo(bazel_lib_toolchains, "yq_linux_ppc64le")
use_repo(bazel_lib_toolchains, "yq_linux_s390x")
use_repo(bazel_lib_toolchains, "yq_windows_amd64")
use_repo(bazel_lib_toolchains, "bsd_tar_toolchains")
```

bullseye.yaml
```
version: 1

sources:
  - channel: bullseye main
    url: https://snapshot-cloudflare.debian.org/archive/debian/20240411T144035Z
  - channel: bullseye-security main
    url: https://snapshot-cloudflare.debian.org/archive/debian-security/20240411T144035Z
  - channel: bullseye-updates main
    url: https://snapshot-cloudflare.debian.org/archive/debian/20240411T144035Z

archs:
  - "amd64"
  - "arm64"

packages:
  - "coreutils"
```

```
% bazel run @bullseye//:lock
[...]

PR #33, opened 2024-04-14
Fix lockfiles in root of workspace

jacobshirley:
Fixes https://github.com/GoogleContainerTools/rules_distroless/issues/31


loosebazooka:
@jacobshirley we require a signed CLA before proceeding

jacobshirley:
@loosebazooka that should be done now

thesayyn (approved):

PR #35, opened 2024-04-17
fix: always install highest version

thesayyn:
This fixes an issue where package resolution did not pick the latest version of a package due to missing logic in the resolution algorithm. 

https://askubuntu.com/questions/1395561/what-is-regarded-as-the-latest-version-of-a-package-by-apt

### DIFF
--- a/WORKSPACE.bazel
+++ b/WORKSPACE.bazel
@@ -35,3 +35,14 @@ deb_index(
 load("@bullseye//:packages.bzl", "bullseye_packages")
 
 bullseye_packages()
+
+# bazel run @apt_security//:lock
+deb_index(
+    name = "apt_security",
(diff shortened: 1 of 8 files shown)


WORKSPACE.bazel
```diff
@@ -35,3 +35,14 @@ deb_index(
 load("@bullseye//:packages.bzl", "bullseye_packages")
 
 bullseye_packages()
+
+# bazel run @apt_security//:lock
+deb_index(
```
loosebazooka, line 40: Should these comments stay in here?

loosebazooka (approved):

WORKSPACE.bazel
thesayyn: yeah those are fine

PR #44, opened 2024-04-30
fix package `:data` visibility so it's compatible with `cacerts()`

lazcamus:
set package `:data` visibility to public so it can be used in a call to `cacerts()`

I stuck a quick check into `examples/apt` tests to verify that it behaves as expected.

### DIFF
--- a/apt/private/index.bzl
+++ b/apt/private/index.bzl
@@ -31,7 +31,8 @@ load("@rules_distroless//distroless:defs.bzl", "flatten")
 flatten(
     name = "data",
     tars = [{src}],
-    compress = "gzip"
+    compress = "gzip",
+    visibility = ["//visibility:public"],
 )
 
 alias(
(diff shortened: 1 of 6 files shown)


lazcamus:
The `bullseye.lock.json` diff has a spaces vs tabs change that was unintentional ... the diff looks sane with whitespace ignored

I'm not sure if this is a byproduct of another change and the lock file never got regenerated, or maybe my environment is somehow creeping in?

thesayyn:
Yeah, we should be ignoring the lockfiles from formatter. Feel free to revert those lockfile changes.

lazcamus:
@thesayyn lockfile is reverted ... though it'll have a legit change if you re-gen it b/c I added a package

thesayyn:
that's true, sorry, you can regen it.

lazcamus:
> that's true, sorry, you can regen it.

regenned the lockfile and rebased on `main` so the PR should be current

oddly, the spurious whitespace diff I was seeing before is no longer occurring 🤷

thesayyn (approved):

PR #46, opened 2024-05-08
fix: evaluate return_code from rctx.execute() call when fetching package indexes

alexconrey:
# Problem
The `rctx.execute()` call when fetching package indexes does not evaluate the tool's return code. This allows for weirdness should an archive be uploaded to an HTTP server which was improperly compressed or contains otherwise invalid data.

# Solution
Evaluate the return code from the executing tools. Both tools return non-zero codes when an error is reached.


thesayyn:
Could you rebase?

alexconrey:
@thesayyn think this is ready for another glance. Sorry for the delay!

thesayyn (approved):

thesayyn:
Thank you!

PR #50, opened 2024-05-19
Bzlmod support

jacobshirley:
Hi there, I'm not sure if this is the best-practice way to implement bzlmod, but it does seem to work. Any feedback welcome. I've created a test for it in e2e/ so that the MODULE.bazel is different to the root one like the examples/ have; but I can move to examples/ if that would be better.


thesayyn:
Hey, it looks close, I'll take a look next week.

e2e/bzlmod/BUILD.bazel
```diff
@@ -0,0 +1,123 @@
+load("@aspect_bazel_lib//lib:tar.bzl", "tar")
```
thesayyn, line 1: Can we add the bzlmod example directly into `e2e/smoke` so that example is both BZLMOD and WORKSPACE?

thesayyn (commented):
This looks great!  A few nits.

jacobshirley:
Those tests are now combined, I had to delete the WORKSPACE.bzlmod file in e2e/smoke because it seems that the dependencies of rules_distroless are managed by the MODULE.bzl, so with that WORKSPACE.bzlmod file it would complain that it could not find certain dependencies (e.g. rules_oci)

apt/private/index.bzl
```diff
@@ -51,13 +52,16 @@ filegroup(
...
+    package_name = rctx.attr.name.split("~")[-1]
+
     package_defs = [
         '"""Generated by rules_distroless. DO NOT EDIT."""',
         """load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")""",
         "",
         "",
         "# buildifier: disable=function-docstring",
-        "def {}_packages():".format(rctx.attr.name),
+        "def {}_packages():".format(package_name),
```
jacobshirley, line 64: I also had to get the last component of the path, as otherwise it would put ~ in the function name and then error: `syntax error at '~': expected (`

thesayyn (approved):
LGTM, thanks!

Issue #36, opened 2024-04-17
FR: `max depth exceeded` when trying to load images that have many apt dependencies


juanzolotoochin:
To repro:

- Add a few larger apt dependencies to `main/examples/apt/bullseye.yaml`. I can repro this with just adding `python3-minimal`
- Remove the `if` statement here: https://github.com/GoogleContainerTools/rules_distroless/blob/789bf1df506c59ccb4e72ae16220101e2148e19f/apt/private/package_resolution.bzl#L105. This is important because it fixes https://github.com/GoogleContainerTools/rules_distroless/issues/29 which is causing a lot of tars to not be added to the image.
- Run `bazel run @bullseye//:lock`
- Add the dependencies to the image target here https://github.com/GoogleContainerTools/rules_distroless/blob/main/examples/apt/BUILD.bazel#L49
- Run the tests: `bazel test //examples/apt/... --test_output=all`

**Expected output:** test passing

**Actual output:** a bunch of tests failing with `Error: error setting env vars: Error creating container: no such image`

From my conversations with @thesayyn , the underlying error is `{"errorDetail":{"message":"max depth exceeded"},"error":"max depth exceeded"}`

Googling for that error, seems to be caused by too many docker layers. One workaround is to combine those layers into a single tar:

```
pkg_tar(
    name = "linux_layers.tar",
    deps = select({
        "@platforms//cpu:arm64": [
            "%s/arm64" % package
            for package in PACKAGES
        ],
        "@platforms//cpu:x86_64": [
            "%s/amd64" % package
            for package in PACKAGES
        ],
    }),
)
```

Then using that `tar` on the image instead of each debian package tar.

Presumably, the same should be achievable using `rules_distroless` `flatten` rule. However this:

```
flatten(
    name = "linux_layers_flatten.tar",
    tars = select({
        "@platforms//cpu:arm64": [
            "%s/arm64" % package
            for package in PACKAGES
        ],
        "@platforms//cpu:x86_64": [
            "%s/amd64" % package
            for package in PACKAGES
        ],
    }),
)
```

does not resolve it. There seems to be an issue with duplicate file paths in `flatten`:

```
$ bazel build //examples/apt:tarball

$ docker load < bazel-bin/examples/apt/tarball/tarball.tar
duplicates of file paths not supported
```

thesayyn:
Thank you for filing this!

lazcamus:
[...]

PR #55, opened 2024-06-03
Support Flat Repository Format

alexconrey:
# Problem
When working with "flat repository format" external repositories, the assumption made by `rules_distroless` is to use the known repository hierarchy within `dists` directory. See: https://wiki.debian.org/DebianRepository/Format#Flat_Repository_Format

In a flat repository, everything is "flat" within the top-level of the repository and as such - the `{}/dists/{}/{}/binary-{}/Packages.{}".format(url, dist, comp, arch, file_type)` logic will always fail to find the Package indices.

Example: https://cloud.r-project.org/bin/linux/debian/bullseye-cran40/

# Solution
Add supporting logic around Package index fetch to:
* Default to the nominal Debian Repository Format, which includes all files in `dists/` or `pool/`
* Fall back to flat repository format if retrievals from `dists/...` default path fail. Uses path provided in `sources.url` YAML for respective repository as base for URL when pulling package files

## Notes
* A flat repository will end up producing a similar WARNING output as a result of these changes: 
```
WARNING: Download from https://cloud.r-project.org/bin/linux/debian/bullseye-cran40/dists/bullseye-cran40//binary-amd64/Packages.xz failed: class java.io.FileNotFoundException GET returned 404 Not Found
WARNING: Download from https://cloud.r-project.org/bin/linux/debian/bullseye-cran40/Packages.xz failed: class java.io.FileNotFoundException GET returned 404 Not Found
WARNING: Download from https://cloud.r-project.org/bin/linux/debian/bullseye-cran40/dists/bullseye-cran40//binary-amd64/Packages.gz failed: class java.io.FileNotFoundException GET returned 404 Not Found
WARNING: Download from https://cloud.r-project.org/bin/linux/debian/bullseye-cran40/dists/bullseye-cran40//binary-arm64/Packages.xz failed: class java.io.FileNotFoundException GET returned 404 Not Found
WARNING: Download from https://cloud.r-project.org/bin/linux/debian/bullseye-cran40/dists/bullseye-cran40//binary-arm64/Packages.gz failed: class java.io.FileNotFoundException GET returned 404 Not Found
```
  * Testing this specific case required adding a package which contains over a dozen dependencies. To ensure accurate testing, I included it in this PR. I think this could go either way with including these tests upstream, this may be a case where the feature is implemented but a warning in a known-working case is less than desirable for contributors making upstream changes
* Specific test package put image over layer depth limits. Implemented workaround noted in: https://github.com/GoogleContainerTools/rules_distroless/issues/36 (`pkg_tar`)
*  `pkg_tar` requires `rules_pkg` inclusion, added to `MODULE.bazel`

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -10,6 +10,7 @@ bazel_dep(name = "bazel_skylib", version = "1.5.0")
 bazel_dep(name = "aspect_bazel_lib", version = "2.7.3")
 bazel_dep(name = "container_structure_test", version = "1.16.0")
 bazel_dep(name = "rules_oci", version = "1.7.4")
+bazel_dep(name = "rules_pkg", version = "0.10.1")
 
 bazel_lib_toolchains = use_extension("@aspect_bazel_lib//lib:extensions.bzl", "toolchains")
 use_repo(bazel_lib_toolchains, "bsd_tar_toolchains")
--- a/apt/private/package_index.bzl
+++ b/apt/private/package_index.bzl
@@ -9,19 +9,35 @@ def _fetch_package_index(rctx, url, dist, comp, arch, integrity):
     r = {"success": False, "integrity": None}
 
     decompression_successful = False
+
     for file_type, tool in file_types.items():
         output = "{}/Packages.{}".format(target_triple, file_type)
-        r = rctx.download(
-            url = "{}/dists/{}/{}/binary-{}/Packages.{}".format(url, dist, comp, arch, file_type),
-            output = output,
-            integrity = integrity,
-            allow_fail = True,
-        )
-        if r.success:
-            re = rctx.execute(tool + [output])
-            if re.return_code == 0:
-                decompression_successful = True
-                break
+
+        # Attempt to pull Packages archive from within the `dists` hierarchy first,
+        # if that fails - assume the repository is of type "flat"
+        try_urls = [
+            "{}/dists/{}/{}/binary-{}/Packages.{}".format(url, dist, comp, arch, file_type),
+            "{}/Packages.{}".format(url, file_type),
+        ]
+
+        for try_url in try_urls:
+            r = rctx.download(
+                url = try_url,
+                output = output,
+                integrity = integrity,
+                allow_fail = True,
+            )
+            if r.success:
+                re = rctx.execute(tool + [output])
+                if re.return_code == 0:
+                    decompression_successful = True
+                    break
+
+        # If we've set `decompression_successful` to true at any point in the iteration loop,
+        # break out of the loop as we have found and extracted the Package archive and have what is
+        # needed to continue.
+        if decompression_successful:
+            break
 
     if not r.success:
         fail("unable to download package index")
--- a/examples/debian_snapshot/BUILD.bazel
+++ b/examples/debian_snapshot/BUILD.bazel
@@ -3,6 +3,7 @@ load("@container_structure_test//:defs.bzl", "container_structure_test")
 load("@rules_distroless//apt:defs.bzl", "dpkg_status")
 load("@rules_distroless//distroless:defs.bzl", "cacerts", "group", "passwd")
 load("@rules_oci//oci:defs.bzl", "oci_image", "oci_tarball")
+load("@rules_pkg//pkg:tar.bzl", "pkg_tar")
 
 passwd(
     name = "passwd",
@@ -65,17 +66,36 @@ PACKAGES = [
     "@bullseye//perl",
 ]
 
+FLAT_REPO_PACKAGES = [
+    "@bullseye//r-base-core",
+]
+
+# Avoids `max depth issue` described here: https://github.com/GoogleContainerTools/rules_distroless/issues/36
+pkg_tar(
+    name = "flat_repo_packages_layer",
+    deps = select({
+        "@platforms//cpu:arm64": [
+            "%s/arm64" % package
+            for package in FLAT_REPO_PACKAGES
+        ],
+        "@platforms//cpu:x86_64": [
+            "%s/amd64" % package
+            for package in FLAT_REPO_PACKAGES
+        ],
+    }),
+)
+
 # Creates /var/lib/dpkg/status with installed package information.
 dpkg_status(
     name = "dpkg_status",
     controls = select({
         "@platforms//cpu:arm64": [
             "%s/arm64:control" % package
-            for package in PACKAGES
+            for package in PACKAGES + FLAT_REPO_PACKAGES
         ],
         "@platforms//cpu:x86_64": [
             "%s/amd64:control" % package
-            for package in PACKAGES
+            for package in PACKAGES + FLAT_REPO_PACKAGES
         ],
     }),
 )
@@ -93,6 +113,7 @@ oci_image(
         ":group",
         ":dpkg_status",
         ":cacerts",
+        ":flat_repo_packages_layer",
     ] + select({
         "@platforms//cpu:arm64": [
             "%s/arm64" % package
--- a/examples/debian_snapshot/bullseye.yaml
+++ b/examples/debian_snapshot/bullseye.yaml
@@ -16,7 +16,8 @@ sources:
     url: https://snapshot-cloudflare.debian.org/archive/debian-security/20240210T223313Z
   - channel: bullseye-updates main
     url: https://snapshot-cloudflare.debian.org/archive/debian/20240210T223313Z/
-
+  - channel: bullseye-cran40
+    url: https://cloud.r-project.org/bin/linux/debian/bullseye-cran40/
 archs:
   - "amd64"
   - "arm64"
@@ -33,3 +34,5 @@ packages:
   - "perl"
   # test cacerts() compatibility
   - "ca-certificates"
+  # test flat repository format package references
+  - "r-base-core"
--- a/examples/debian_snapshot/test_linux_amd64.yaml
+++ b/examples/debian_snapshot/test_linux_amd64.yaml
@@ -18,6 +18,7 @@ commandTests:
       - ncurses-base/now 6\.2\+20201114-2\+deb11u2 all \[installed,local\]
       - perl/now 5\.32\.1-4\+deb11u3 amd64 \[installed,local\]
       - tzdata/now 2024a-0\+deb11u1 all \[installed,local\]
+      - r-base-core/now 4.4.0-2~bullseyecran.0 i386 \[installed,local\]
   - name: "whoami"
     command: "whoami"
     expectedOutput: [r00t]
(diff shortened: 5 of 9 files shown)


MODULE.bazel
```diff
@@ -10,6 +10,7 @@ bazel_dep(name = "bazel_skylib", version = "1.5.0")
 bazel_dep(name = "aspect_bazel_lib", version = "2.7.3")
 bazel_dep(name = "container_structure_test", version = "1.16.0")
 bazel_dep(name = "rules_oci", version = "1.7.4")
+bazel_dep(name = "rules_pkg", version = "0.10.1")
```
thesayyn, line 13: can we not add a dependency here?

apt/private/package_index.bzl
```diff
@@ -9,19 +9,35 @@ def _fetch_package_index(rctx, url, dist, comp, arch, integrity):
...
-            re = rctx.execute(tool + [output])
-            if re.return_code == 0:
-                decompression_successful = True
-                break
+
+        # Attempt to pull Packages archive from within the `dists` hierarchy first,
+        # if that fails - assume the repository is of type "flat"
+        try_urls = [
+            "{}/dists/{}/{}/binary-{}/Packages.{}".format(url, dist, comp, arch, file_type),
+            "{}/Packages.{}".format(url, file_type),
```
thesayyn, line 20: How does flat repositories play with different architectures?  in the manifest format we support specifying architectures matrix which i don't see being applied here. 

I believe we should disallow flat repositories to be combined with non flat repositories to prevent confusion.

ericlchen1:
Is this still being worked on? If not, I'll take a look at this and probably make a separate PR since we also need this functionality (preferably without the patch we are using).

apt/private/package_index.bzl
jjmaestro: I don't think disallowing flat repos mixed with "classic" repos is really needed.

You can have flat repos with multiple architectures (e.g. https://cloud.r-project.org/bin/linux/debian/bullseye-cran40/ has `i386` and `amd64` packages). The NVIDIA repos are "a bit weirder" since they are flat and have multiple architectures (e.g. https://developer.download.nvidia.com/compute/cuda/repos/debian12/x86_64/ has `i386` and `amd64` packages) but they also seem to have separate flat repos for "other architectures" (e.g. https://developer.download.nvidia.com/compute/cuda/repos/debian12/sbsa/ has `arm64` packages).

Personally, as I mentioned in https://github.com/GoogleContainerTools/rules_distroless/pull/64#issuecomment-2340956897 if there are specific packages in specific architectures I'd just separate them and use another manifest.

apt/private/package_index.bzl
```diff
@@ -9,19 +9,35 @@ def _fetch_package_index(rctx, url, dist, comp, arch, integrity):
...
-            allow_fail = True,
-        )
-        if r.success:
-            re = rctx.execute(tool + [output])
-            if re.return_code == 0:
-                decompression_successful = True
-                break
+
+        # Attempt to pull Packages archive from within the `dists` hierarchy first,
+        # if that fails - assume the repository is of type "flat"
```
jjmaestro, line 17: This is not following the Debian repo format spec that you mentioned. In there, they show the two `sources.list` formats, one for "classic" repos:
```
deb uri distribution [component1] [component2] [...]
```
and one for flat repos:
```
deb uri directory/
```

I think the convention is that `directory` always ends with the slash `/` and, while `distribution` can have slashes, it never ends with a slash.

IMHO we should use the same in the YAML manifest, something like:
```
version: 1

sources:
  - channel: bullseye main
    url: https://snapshot-cloudflare.debian.org/archive/debian/20240210T223313Z
  - channel: bullseye-cran40/
    url: https://cloud.r-project.org/bin/linux/debian

archs:
  - amd64

packages:
  - bash
  - r-base-core
```
and use the slash to distinguish between "classic" and flat repos.

Also, although the `sources.list` spec doesn't seem to require `components`, we are also parsing them so I guess that's another difference that should be made in the parsing of the manifest (and validation).

examples/debian_snapshot/bullseye.yaml
```diff
@@ -33,3 +34,5 @@ packages:
   - "perl"
   # test cacerts() compatibility
   - "ca-certificates"
+  # test flat repository format package references
+  - "r-base-core"
```
jjmaestro, line 38: I think the issue with the `max depth exceeded` is not the only problem of using this package / repo for testing.

It seems that the versions move quite fast... I've tried running the tests of this PR and they initially failed because the version in the lock file didn't exist, so I had to update the lock. Ideally, the tests should be done with a snapshot repo that will "always" exist.

Also,, the R-project repo doesn't have `arm64` packages (it only has `i386` and `amd64`, and a few `all`). Thus, the tests will fail for `arm64`. I think it would be best to separate the test into another manifest file and another `BUILD` that e.g. only uses that package and only runs for `amd64`.

Or, preferably, find another flat repo that also has a snapshot... probably not an easy task! 😅 If I come up with such a repo, I'll comment here as well.

apt/private/package_index.bzl
thesayyn: i agree with your findings there.

apt/private/package_index.bzl
thesayyn: Hmm, that makes this even more confusing. How does flat repos interact with arches? for instance if you want to install `i386` and `amd64` of a package from   https://developer.download.nvidia.com/compute/cuda/repos/debian12/x86_64/ ? 

I don't see how this PR allows discovery of arches in flat repositories as of today.

jjmaestro: @thesayyn ah, maybe you are right and there's more fixes needed. I'll try to test the PR and see where it breaks.

examples/debian_snapshot/test_linux_amd64.yaml
```diff
@@ -18,6 +18,7 @@ commandTests:
       - ncurses-base/now 6\.2\+20201114-2\+deb11u2 all \[installed,local\]
       - perl/now 5\.32\.1-4\+deb11u3 amd64 \[installed,local\]
       - tzdata/now 2024a-0\+deb11u1 all \[installed,local\]
+      - r-base-core/now 4.4.0-2~bullseyecran.0 i386 \[installed,local\]
```
jjmaestro, line 21: `i386` is the wrong architecture here, this is caused by a bug in `_parse_package_index`, see my comments before in this commit, in [`apt/private/package_index.bzl`](https://github.com/GoogleContainerTools/rules_distroless/pull/55/files/b84a604cd93533afd2c222cca03f4f63a35df85d#diff-2c0a36dc8512cca883457e90c6bd82fb340b10d91d4f0127eb9e5aae38e68b11).

apt/private/package_index.bzl
jjmaestro: @thesayyn OK, I think I got it, the main issue that I found so far (and it's been a very quick look so maybe there's more issues / corner cases) is that in `_parse_package_index` we force `arch` to be the one we are using in the for-loop so that's a problem for `Package` indexes that contain multiple architectures as we will get the package info from a different arch.

This is actually happening in this PR and the proof is in the lock and, easier to see, in the test itself :D [Check my comment in `examples/debian_snapshot/test_linux_amd64.yaml`](https://github.com/GoogleContainerTools/rules_distroless/pull/55/commits/b84a604cd93533afd2c222cca03f4f63a35df85d#diff-1558de2257da2042402f3f76a361d4e7f778d60ca44673dcc0f78d8832bd3666), the new `expectedOutput` for the `r-base-core` package is set as `i386` instead of the correct `amd64` precisely because the current code in this PR is selecting the wrong package architecture:
```
(...)
      - r-base-core/now 4.4.1-1~bullseyecran.0 i386 \[installed,local\]
```

Now, the fix that I would propose here is the following:
```diff
diff --git a/apt/private/package_index.bzl b/apt/private/package_index.bzl
index 34f8202..8d6f922 100644
--- a/apt/private/package_index.bzl
+++ b/apt/private/package_index.bzl
@@ -77,7 +77,8 @@ def _parse_package_index(state, contents, arch, root):
 
         if len(pkg.keys()) != 0:
             pkg["Root"] = root
-            util.set_dict(state.packages, value = pkg, keys = (arch, pkg["Package"], pkg["Version"]))
+            arch_ = arch if pkg["Architecture"] == "all" else pkg["Architecture"]
+            util.set_dict(state.packages, value = pkg, keys = (arch_, pkg["Package"], pkg["Version"]))
             last_key = ""
             pkg = {}
```

IMHO this "is not great" because it's duping the packages that are valid across all architectures N times (as many as architectures we have in the manifest). It's not too bad, it's just some storage space wasted and "some mismatch" between the repo structure that is on disk (the per-arch folders) VS the actual contents of the tar files (this can be seen in the dpk-status in the test, e.g. `tzdata` is also `all` but the folder and the "rules_distroless arch" for the package is "amd64" or whatever other arch we are using). But... I don't see how to easily fix this without major changes (and even then I don't know how I'd do it).

examples/debian_snapshot/BUILD.bazel
```diff
@@ -65,17 +66,36 @@ PACKAGES = [
...
+            "%s/arm64" % package
+            for package in FLAT_REPO_PACKAGES
+        ],
+        "@platforms//cpu:x86_64": [
+            "%s/amd64" % package
+            for package in FLAT_REPO_PACKAGES
+        ],
+    }),
+)
+
```
jjmaestro, line 87: Found a way to easily avoid having to flatten stuff here in the test and depend on `rules_pkg`, etc.

You can simply use the `:data` target of the package so that you avoid pulling in all of the dependencies :)  Sure, that's "broken" for a real-life image but it'll work to show that the specific package is installed and thus it can be used to validate the test without having to bring in more dependencies and potentially have slower tests, etc.

jjmaestro:
@alexconrey there's quite a few changes to make... IMHO from "first to last":

* separate test to avoid using `arm64` since it's not a supported architecture for the `r-project.org` repo
* fix the bug in `_parse_package_index`
* detect flat VS "canonical" repos in the manifest and avoid the `404` warnings
* nit: to avoid the `max depth` issue and to avoid using `rules_pkg` and/or flattening large amounts of dependencies, we can use the `:data` part of `r-base-core` in the test.

At this point, and given that there's been a few attempts at adding support for flat repos (this PR plus #64 and #67 that I could find) I'm a bit wary of doing yet-another-PR... but I'm also happy to do it and get it done ASAP.

@thesayyn what do you think? would it be OK to create "one 💍 PR to merge them all"? 😅

jjmaestro:
OK, so I finally did try to write a PR that adds support for multi-arch and single-arch flat repos and addresses the issue with the NVIDIA CUDA flat repos. Hopefully that PR superseeds the others and can be merged and 1 + 3 PRs closed 😄 

@alexconrey please have a look and let me know what you think and if #86 works for you. Happy to address any issues or concerns if there's something that breaks in your use-case!

thesayyn:
This PR has stalled, and needs some proper design before we can accept a PR for it. I am going to close this, thank you for the effort.

PR #57, opened 2024-06-16
Update README.md

alexeagle:
Fix bad markup. Why is there HTML in here anyhow??


thesayyn (approved):

thesayyn:
> Why is there HTML in here anyhow

probably because my markdown skills are bad, was there an alternative?

PR #59, opened 2024-06-26
feat: support optional dependencies

### LINKED ISSUE #27: FR: Support optional dependencies
Currently, we skip optional dependencies in transitive closure resolution.

thesayyn:
WIP, fixes https://github.com/GoogleContainerTools/rules_distroless/issues/27


loosebazooka:
How do optionals work?
If I see an optional dep -- but it's included by someone else as non-optional then I resolve it?

thesayyn:
> If I see an optional dep -- but it's included by someone else as non-optional then I resolve it

optional is misnomer by me, it basically means,  either one of these packages are required.

alexeagle (approved):

PR #63, opened 2024-07-09
fix: auto-setting lock attribute on deb_index

alexeagle:
There were three bugs:
- Need to check which WORKSPACE/WORKSPACE.bazel file to update
- Under bzlmod, printed the canonical name (with a bunch of tilde) rather than apparent name of the rule
- Missed escaping for quotes, so the command we printed for users to run had too many arguments

Also as a convenience, check if the user setup `tools/buildozer` so we don't rely on the $PATH if we can help it.

### DIFF
--- a/apt/private/resolve.bzl
+++ b/apt/private/resolve.bzl
@@ -17,15 +17,33 @@ echo ''
 echo 'Writing lockfile to {workspace_relative_path}' 
 cp $lock {workspace_relative_path}
 
+# Detect which file we wish the user to edit
+if [ -e $BUILD_WORKSPACE_DIRECTORY/WORKSPACE ]; then
+    wksp_file=WORKSPACE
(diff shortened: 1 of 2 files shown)


apt/private/resolve.bzl
```diff
@@ -17,15 +17,33 @@ echo ''
 echo 'Writing lockfile to {workspace_relative_path}' 
 cp $lock {workspace_relative_path}
 
+# Detect which file we wish the user to edit
+if [ -e $BUILD_WORKSPACE_DIRECTORY/WORKSPACE ]; then
```
thesayyn, line 21: looks like we should also support MODULE.bazel here.

thesayyn (approved):

apt/private/resolve.bzl
alexeagle: yes, will file a follow-up

Issue #66, opened 2024-07-18
FR: Problems using deb_index to install CUDA packages


benmccown:
We've been trying to use `deb_index` to install the necessary CUDA packages that we ultimately need pulled into a rules_oci container image, where we're using Ubuntu 20.04 as our base.

We ran into a few different problems getting this to work since [Nvidia's debian package repo](https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/) doesn't adhere to the same principles as Ubuntu's package repo. For example, Ubuntu's `Packages` manifest for the apt repo is located at a very specific URL which is hardcoded here in rules_distroless. In Nvidia's case, the `Packages` manifest sits at the same URL as the packages themselves. So `deb_index` was incapable of finding the `Packages` manifest. We also ran into an issue where rules_distroless didn't support the `Packages` file being gzipped.

We also saw in `_deb_package_index_impl` that duplicate transitive dependencies weren't handled correctly which caused issues.

At this point these issues we ran into were about two months ago. I forked the project, fixed our issues, and have just now gotten back around to looking at my fork. I can patch rules_distroless using diff based patching in our team's project and move on, but I thought I'd open an issue and PR in case someone else wants to install CUDA packages and has similar problems.

thesayyn:
fixed by https://github.com/bazel-contrib/rules_distroless/pull/222

PR #64, opened 2024-07-11
Add flat repository support for debian images

ericlchen1:
Fixes #56 and followup to https://github.com/GoogleContainerTools/rules_distroless/pull/55

## Problem:
When working with "flat repository format" external repositories, the assumption made by rules_distroless is to use the known repository hierarchy within dists directory. See: https://wiki.debian.org/DebianRepository/Format#Flat_Repository_Format

In a flat repository, everything is "flat" within the top-level of the repository and as such - the {}/dists/{}/{}/binary-{}/Packages.{}".format(url, dist, comp, arch, file_type) logic will always fail to find the Package indices.

Example: https://cloud.r-project.org/bin/linux/debian/bullseye-cran40/ or https://developer.download.nvidia.com/compute/cuda/repos/debian11/x86_64/ which are both flat repositories with arch specific packages (amd64/i386) 

## Solution:
Manifest:
`sources`: add `flat_repository` and `arch` keys
- flat_repository: default false, boolean that represents if source is a flat_repository
- arch: lock repository to a specific architecture. Other architectures will not be considered for lockfile if this is set.

Add `arch_specific_packages` section with format
```
arch_specific_packages:
  architecture:
    - "package1"
    - "package2"
```

PR #67, opened 2024-07-18
Fix issues with deb_index and Nvidia's apt repository

benmccown:
I opened https://github.com/GoogleContainerTools/rules_distroless/issues/66 detailing some of the problems we have run into when trying to use `deb_index` to install Nvidia's CUDA packages from their apt repository.

These fixes are my first pass at resolving issues that were present with the apt package implementation here in rules_distroless.

I did some research to try and find specification as to how `Packages` works with apt and apt-cache and whether there's a specific location that `Packages` should live, but I couldn't find anything detailing a correct implementation for a remote apt package repository. Based on this, it seems to me that [Nvidia's package repository](https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/) is formatted in a valid way and rules_distroless should ideally support it.

I'm fully expecting to need a code revision here, just wanted to share my first iteration to get feedback and see what maintainers want changed.


apt/private/index.bzl
```diff
@@ -85,17 +85,19 @@ def _deb_package_index_impl(rctx):
             ),
         )
 
+        deps = depset([
```
benmccown, line 88: We had packages that shared duplicate transitive dependencies. The duplicate deps were a blocking issue, so I used a depset to dedupe packages.

apt/private/lockfile.bzl
```diff
@@ -16,11 +16,14 @@ def _add_package(lock, package, arch):
     k = _package_key(package, arch)
     if k in lock.fast_package_lookup:
         return
+    filename = package["Filename"]
+    if filename.startswith("./"):
```
benmccown, line 20: When `Packages` lives in the same directory/URL that the packages themselves live, the manifest has `./` as the leading file path for all the packages listed, however we don't want a `./` in the URL.

apt/private/package_index.bzl
```diff
@@ -3,25 +3,36 @@
...
 
 def _fetch_package_index(rctx, url, dist, comp, arch, integrity):
-    target_triple = "{dist}/{comp}/{arch}".format(dist = dist, comp = comp, arch = arch)
+    # Split the URL by the '://' delimiter
+    protocol, rest = url.split("://")
+
+    # Split the rest of the URL by the '/' delimiter and take the first part
+    domain = rest.split("/")[0]
+
+    target_triple = "{domain}/{dist}/{comp}/{arch}".format(domain = domain, dist = dist, comp = comp, arch = arch)
```
benmccown, line 12: We needed to add `domain` to the `target_triple` because dist/comp/arch was duplicated across repositories since Ubuntu and Nvidia's repositories were both `focal/main/x86_64`.

`target_triple` is probably no longer the right term to use. I can change it as you see fit.

apt/private/package_index.bzl
```diff
@@ -3,25 +3,36 @@
...
-            allow_fail = True,
-        )
-        if r.success:
-            re = rctx.execute(tool + [output])
-            if re.return_code == 0:
-                decompression_successful = True
-                break
+        urls = [
+            "{}/dists/{}/{}/binary-{}/Packages.{}".format(url, dist, comp, arch, file_type),
+            "{}/Packages.{}".format(url, file_type),
```
benmccown, line 22: This line adds support for Nvidia's `Packages` location. We had to add gzip support as well since they didn't have an `xz` compressed archive.

benmccown (commented):
Left some comments explaining a few things.

benmccown:
I'm fully expecting to need a code revision here, just wanted to share my first iteration to get feedback and see what maintainers want changed.

benmccown:
@alexeagle @thesayyn Just tagging y'all as you look like recently active maintainers.

benmccown:
@alexeagle @thesayyn bump. Any thoughts on what I'm trying to accomplish here? Nvidia's cuda packages (and corresponding apt repository) seem like common packages folks might need to be able to pull.

thesayyn:
This is same as https://github.com/GoogleContainerTools/rules_distroless/pull/64

apt/private/lockfile.bzl
jjmaestro: FYI, this is not what the spec says, check https://wiki.debian.org/DebianRepository/Format#Filename:

> Filename
> 
> The mandatory Filename field shall list the path of the package archive relative to the base directory of the repository. The path should be in canonical form, that is, without any components denoting the current or parent directory ("." or ".."). It also should not make use of any protocol-specific components, such as URL-encoded parameters.

I've checked the R-Project flat repos and this doesn't happen / is not needed. However, this is something that happens in the NVIDIA CUDA flat repos . Thus, I think that this it's a quirk of some specific flat repos that don't conform with the Debian repo spec.

apt/private/index.bzl
jjmaestro: Do you have an example the package(s) that broke? Were they packages in the NVIDIA CUDA repos?

jjmaestro:
OK, so I finally did try to write a PR that adds support for multi-arch and single-arch flat repos and addresses the issue with the NVIDIA CUDA flat repos. Hopefully that PR superseeds the others and can be merged and 1 + 3 PRs closed 😄 

@benmccown please have a look and let me know what you think and if #86 works for you. Happy to address any issues or concerns if there's something that breaks in your use-case!

apt/private/lockfile.bzl
benmccown: Super interesting. I was looking for documentation on the manifest format for a debian repository and was having a hard time finding anything authoritative. Thanks for sharing the link.

apt/private/index.bzl
benmccown: @jjmaestro I had this in my sources:

```
  - channel: nvidia cuda
    url: https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64
```

And this in my packages:

```
  # CUDA Dependencies
  - cuda-11-8
  - cudnn9-cuda-11-8
```

I think those would replicate the issue.

jjmaestro: @benmccown OK, I looked more carefully and I was using the wrong `Packages` index / arch but that's also the problem for the repro, I can't get both of those packages for both `amd64` and `arm64` (they are only available for `amd64` IIRC).

Have a look at https://github.com/jjmaestro/rules_distroless/tree/testing-cuda-packages, I have one commit there so far: https://github.com/jjmaestro/rules_distroless/commit/f216178ce98ba1d21c8759818b2aa3949e37cc92, this is a test running on top of https://github.com/GoogleContainerTools/rules_distroless/pull/97.

I've gotten it to work with just the `cuda-11-8` package in `arm64`, and since that's the largest package by far, I think it's probably working but I'll change the example I'm working on to only `amd64` and see if something breaks when adding both packages.

jjmaestro: I've found yet-another-issue when testing with `amd64`... the failure I'm getting is because there's a Debian package that's already compressed as a `tar.gz`, see #98. I'll see if I can work around that issue and test things further but so far, I haven't seen duplicate transitive dependencies.

jjmaestro: Alright, @benmccown please check [jjmaestro/rules_distroless/tree/testing-cuda-packages](https://github.com/jjmaestro/rules_distroless/tree/testing-cuda-packages).

I didn't get any errors / issues with duplicate transitive dependencies. I have it all working after porting #99 and also making sure I flatten the layers to avoid `max depth exceeded` (#36) and also, using `pkg_tar` to flatten, to avoid `duplicates of file paths not supported` Docker errors (e.g. https://github.com/bazelbuild/rules_docker/issues/246).

Can you check that branch and confirm if that works for you?

Thanks!

PR #70, opened 2024-08-08
fix: apt.deb_index should work without a lock

thesayyn:
Fixes the issue where if no `lock` file is passed to `apt.deb_index` it fails horribly.  This change fixes the issue.
Sorry for the big diff because i had to refactor a lot to make it work with bzlmod!


examples/debian_shared_dependencies/BUILD.bazel
```diff
@@ -92,7 +92,7 @@ oci_image(
     }),
 )
 
-oci_tarball(
+oci_load(
```
alexeagle, line 95: nit: is this used anywhere?

e2e/smoke/MODULE.bazel
```diff
@@ -1,22 +1,26 @@
...
 
 apt = use_extension("@rules_distroless//apt:extensions.bzl", "apt")
-apt.deb_index(
+apt.install(
     name = "bullseye",
     lock = ":bullseye.lock.json",
     manifest = ":bullseye.yaml",
 )
+apt.install(
+    name = "bullseye_nolock",
```
alexeagle, line 20: nit: maybe comment about why this is needed?

apt/private/resolve.bzl
```diff
@@ -61,8 +19,11 @@ def _parse_manifest(rctx):
 
     return json.decode(result.stdout if result.stdout != "null" else "{}")
 
-def _deb_resolve_impl(rctx):
-    manifest = _parse_manifest(rctx)
+# This function is shared between BZLMOD and WORKSPACE implementations.
+# INTERNAL: DON NOT DEPEND!
```
alexeagle, line 23: nit: spelling

apt/private/deb_import.bzl
```diff
@@ -0,0 +1,29 @@
+"deb_import"
+
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+# BUILD.bazel template
```
alexeagle, line 5: nit: maybe it should go in `BUILD.deb` following typical convention?

alexeagle (approved):

examples/debian_shared_dependencies/BUILD.bazel
thesayyn: not by our tests no, but there for the completeness of the example

e2e/smoke/MODULE.bazel
thesayyn: Sure, the doc here explains why. i will leave comment there for visibility. https://github.com/GoogleContainerTools/rules_distroless/blob/main/docs/apt.md#lockfile-generation

apt/private/deb_import.bzl
thesayyn: Hmm, it could go into a file and passed to http_archive build_file attribute. I'll reconsider it in the future.

PR #75, opened 2024-08-17
tar: Make all gzip compression hermetic

DolceTriade:
We weren't passing the flag to make gzip hermetic. Without the flag to
ignore timestamp, the shasum will differ based on the system time.

Also, DEFAULT_ARGS contained "--options=gzip:...", which is invalid when
using any other compression scheme. Therefore, remove that option from
DEFAULT_ARGS and provide an alternate function called
`add_default_compression_args` to add per-compression type default
arguments which can do things like make the compression more hermetic.

Repro instructions:
```
bazel test //...
find -L bazel-bin -name "*.gz" -exec sha256sum {} \; > sha
bazel clean --expunge
bazel test //...
sha256sum -c sha
```

Before, the above test would fail with sha256sum mismatches. With this patch, all shasums match.

### DIFF
--- a/distroless/private/cacerts.bzl
+++ b/distroless/private/cacerts.bzl
@@ -43,7 +43,6 @@ oci_image(
 ```
 """
 
-
 def _cacerts_impl(ctx):
     bsdtar = ctx.toolchains[tar_lib.TOOLCHAIN_TYPE]
 
--- a/distroless/private/flatten.bzl
+++ b/distroless/private/flatten.bzl
@@ -11,8 +11,10 @@ def _flatten_impl(ctx):
     output = ctx.actions.declare_file(ctx.attr.name + ext)
 
     args = ctx.actions.args()
+    args.add_all(tar_lib.DEFAULT_ARGS)
     args.add("--create")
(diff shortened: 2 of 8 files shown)


thesayyn (approved):
Thanks!

DolceTriade:
> tar: Unknown module name: `gzip'

Hm, I thought that the aspect build tar would support this universally?

DolceTriade:
Ok, tests pass now. While running I realized a lot more than flatten was not hermetic, so I fixed those as well. Please let me know if you'd like a different approach. I also updated the commit message with the motivations.

jjmaestro:
I'd love to get this merged, I'm also using `flatten` and will need the hermeticity! :)

thesayyn:
@DolceTriade could you rebase and solve the merge conflicts? there are also some buildifier issues it seems

DolceTriade:
Done. @thesayyn PTAL, thanks!

thesayyn (approved):

DolceTriade:
I'm guessing the buildifier failures are due to differing versions of buildifier...

jjmaestro:
Yeah, the buildifier errors are due to other PRs, I've fixed it in #108

PR #88, opened 2024-09-19
feat: add MODULE.bazel support to copy.sh plus silent lockless repos

jjmaestro:
> [!NOTE]  
>  Stacked on top of #87

## feat: add support for MODULE.bazel to the lock copy.sh script

Add support for `MODULE.bazel` to the lock script and avoid printing an unnecessary (and annoying 😅) error when building in a "modern repo".

## feat: avoid DEBUG messages for lockless repos
Add a `nolock` attribute to avoid getting annoying `DEBUG` messages for repos that we explicitly want to run without a lock.


jjmaestro:
@thesayyn done! I've also rebased the rest

thesayyn (approved):

PR #89, opened 2024-09-19
refactor: MEGA refactor

jjmaestro:
<PLACEHOLDER>, PR description will be a combination of all of the commits, but I wanted to get a draft PR first.


thesayyn:
Please, create these as individual PRs. We won't be reviewing PR as this big.

jjmaestro:
@thesayyn ok, will do!

But just FYI I thought it would actually be much easier to do it this way since I've already separated the small fixes in #87, the small features in #88, then all of #89 is literally refactoring without adding any functionality, and then #90 has the flat repo feature + NVIDIA fix.

In here each commit is small enough and "self-contained", aiming to clean or improve a specific part and has a proper commit message explaining everything done in it (or so I hope!) and thus it can be easily reviewed one-by-one in the PR.

Anyway, I'll start breaking it up then!

PR #92, opened 2024-09-19
refactor: improve copy sh

jjmaestro:
> [!NOTE]  
>  Stacked on top of #91 

## refactor: improve the lock copy.sh script
* separate the script into a template file so it's easier to shellcheck and syntax highlight in editors.

* shellcheck the script and remove all SC2086 warnings ("Double quote to prevent globbing and word splitting")

* improve the buildozer help messages in the copy.sh script:
  * reduce duplication of buildozer command
  * add a more clear autofix bazel run command that can be easily copy-pasted

* change some of the variable names in the copy.sh template for longer, easier to understand names (repo_name >> name, lock_label >> label)

* move repo_name and workspace_relative_path into variables to reduce line length and improve readability


jjmaestro:
@thesayyn could you continue reviewing the stack of PRs? 😅 🙏  this one and the previous 3 PRs in the stack (87, 88, and 91) should be very easy to review, they add a few things before the larger refactors.

Thanks!!

apt/private/resolve.bzl
```diff
@@ -149,16 +105,22 @@ def _deb_resolve_impl(rctx):
...
+    lock_label = rctx.attr.manifest.relative(lock_filename)
+    workspace_relative_path = "{}{}".format(
+        ("%s/" % lock_label.package) if lock_label.package else "",
+        lock_label.name,
+    )
+
     rctx.file(
         "copy.sh",
-        _COPY_SH_TMPL.format(
+        rctx.read(Label("//apt/private:copy.sh.tmpl")).format(
```
thesayyn, line 117: This lead to repository restarts, which is not a good thing. can you revert this back to be a static string here?

thesayyn: If we want this to be subject to shellcheck, then you should create a private attribute to this repository rule and use this label as the default value to prevent repository rule restarts.

jjmaestro: Oh wow! I don't even understand what you are saying 😅 I'm quite a Bazel n00b so I don't know what a "repo restart" is. You mean the `Label` in there  causes "cache problems" / somehow makes the build "not immutable" or something and it causes a rebuild or some other side-effect!? :-?  I'd love to learn more about this!

Re. creating a private attribute, would something like this work?
```starlark
def _deb_resolve_impl(rctx):
    lockf = internal_resolve(rctx, rctx.attr.yq_toolchain_prefix, rctx.attr.manifest, rctx.attr.resolve_transitive)
    lockf.write("lock.json")

    lock_filename = rctx.attr.manifest.name.replace(".yaml", ".lock.json")
    lock_label = rctx.attr.manifest.relative(lock_filename)
    workspace_relative_path = "{}{}".format(
        ("%s/" % lock_label.package) if lock_label.package else "",
        lock_label.name,
    )

    rctx.file(
        "copy.sh",
        rctx.read(rctx._copy_sh_tmpl).format(
            # NOTE: the split("~") is needed when we run bazel from another
            # directory, e.g. when running e2e tests we change dir to e2e/smoke
            # and then rctx.name is 'rules_distroless~~apt~bullseye'
            repo_name = rctx.name.split("~")[-1].replace("_resolve", ""),
            lock_label = lock_label,
            workspace_relative_path = workspace_relative_path,
        ),
        executable = True,
    )

    rctx.file("BUILD.bazel", _BUILD_TMPL)

deb_resolve = repository_rule(
    implementation = _deb_resolve_impl,
    attrs = {
        "manifest": attr.label(),
        "resolve_transitive": attr.bool(default = True),
        "yq_toolchain_prefix": attr.string(default = "yq"),
        "_copy_sh_tmpl": attr.label(
            defaul="//apt/private:copy.sh.tmpl",
            doc = "INTERNAL, DO NOT USE - " +
                  "private attribute label to prevent repo restart",
        ),
    },
)
```
Thanks!

jjmaestro: Also, re. shellcheck, it won't work straight on the template "obviously" because of the templated variables... it's just much easier to do the manual work of copying the template, changing those variables and running `shellcheck`. Plus IMO it makes the Bazel code / the file much nicer and easier to read 😄  But I'm also happy to put the script back into a string if it's easier / prevents other issues that I don't fully grasp 😅

jjmaestro: Ah, I've asked chatGPT 😆😆

> By using `rctx.read(Label("//apt/private:copy.sh.tmpl"))`, Bazel will re-read the `copy.sh.tmpl` file every time the repository rule runs, which can trigger repository restarts. Repository restarts happen when Bazel notices a change in an external repository, which can lead to performance issues because Bazel will re-run the repository rule and reload dependencies, **even if the file hasn't actually changed**.
>
> By using a private attribute with a default value (like `Label("//apt/private:copy.sh.tmpl")`), you give the rule an explicit, static dependency on this file. This prevents Bazel from thinking the file might change during rule execution, avoiding unnecessary repository restarts.

Let me know if this is correct! :)  I wouldn't have thought that it can make a difference whether the `Label` is an attribute or not 😄  TIL!

jjmaestro:
@thesayyn I think I've addressed the issue in your review re. the `Label` as a private attribute, please have a look and let me know if this is better! And approve the workflow checks please! 🙏 😅

thesayyn (approved):

PR #93, opened 2024-09-19
refactor: package_index

jjmaestro:
> [!NOTE]  
>  Stacked on top of #92

### refactor: `apt_dep_resolver_test`
Cleanup `apt_dep_resolver_test` removing the `apt_deb_repository mock`, it's not really needed once we have proper testing and mocking of `apt_deb_repository` that we can reuse.

### test: `apt_deb_repository`
* Add testing for apt_deb_repository mocking the external / side effects (downloads, decompression, etc).
* Rename _parse_repository back to _parse_package_index

### refactor: cleanup `_resolve_package` `virtual_package` logic
* Refactor the `virtual_package` loop moving `if version` to the outside

### feat: add `parse_provides`
* Add a `parse_provides` method which parses and validates 'Provides'

### refactor: `nested_dict` `struct`
* Add a helper `nested_dict` `struct` that encapsulates the get-set logic
* Remove all the (now) unnecessary wrappers like `_package_versions`, `_virtual_packages`, etc.
* Fix `get()` method to return default_value if no keys are given
* Add `add()` to `nested_dict` `struct` so that we can simplify the virtual package logic in `apt_deb_repository.bzl` `_add_package`

### refactor: `_resolve_all`
* Add assertions for unresolved_deps in apt_dep_resolver_test
* Rename `resolve_all` to `resolve`
* Reorder `(name, version, arch)` args to match the order of the index keys `(arch, name, version)`
* Break up the long lines in comments
* Improve `_ITERATION_MAX_` fail message

### refactor: `_parse_version_constraint`
Cleanup the version constraint parsing using the `VERSION_OPERATORS` that we now have in the version `struct` plus also validating the version being parsed.

### chore: mv resolution_test to `apt_dep_resolver_test`
Tests should match the file they are testing

### chore: move `version_constraint` tests to their own test file
    
`version_constraint.bzl` is a bunch of utils for package version dependency. Either it has enough entity to stand by itself or the functionality should be folded into `version.bzl` and `package_index.bzl` and/or `package_resolution.bzl`.

### refactor: `_version_relop`
Refactor `_version_relop` into a compare method in `version.bzl` plus a `VERSION_OPERATORS` `dict` so that (1) we use the operator strings everywhere and (2) we can use the keys to validate the operators.

### refactor: _fetch_package_index
    
While working with some flaky mirrors and trying to figure out why they were failing I found the `_fetch_package_index` code a bit hard to follow so here's my attempt at streamlining it a bit:
    
* Change the general flow of the for-loop so that we can directly set the reasons for failure in failed_attempts
* Remove both integrity as an argument and as a return value since neither is ever used.
* return the content of the Packages index instead of the path so that (1) we don't need `rctx` anywhere else and (2) is easier to mock.
* Reword failure messages adding more context and debug information
* Shorter lines and templated strings, trying to make the code easier to read and follow.

### refactor: apt_dep_resolver
* Remove unnecessary `state` `struct`, just pass the `repository` `struct` to the resolver methods
* Remove unused `resolve_package` from the public API

### DIFF
--- a/apt/private/apt_deb_repository.bzl
+++ b/apt/private/apt_deb_repository.bzl
@@ -3,9 +3,7 @@
 load(":util.bzl", "util")
 load(":version_constraint.bzl", "version_constraint")
 
-def _fetch_package_index(rctx, url, dist, comp, arch, integrity):
-    target_triple = "{dist}/{comp}/{arch}".format(dist = dist, comp = comp, arch = arch)
-
+def _fetch_package_index(rctx, url, dist, comp, arch):
     # See https://linux.die.net/man/1/xz , https://linux.die.net/man/1/gzip , and https://linux.die.net/man/1/bzip2
     #  --keep       -> keep the original file (Bazel might be still committing the output to the cache)
     #  --force      -> overwrite the output if it exists
@@ -15,47 +13,72 @@ def _fetch_package_index(rctx, url, dist, comp, arch, integrity):
         ".xz": ["xz", "--decompress", "--keep", "--force"],
         ".gz": ["gzip", "--decompress", "--keep", "--force"],
         ".bz2": ["bzip2", "--decompress", "--keep", "--force"],
-        "": ["true"],
+        "": None,
     }
 
     failed_attempts = []
 
-    for (ext, cmd) in supported_extensions.items():
-        output = "{}/Packages{}".format(target_triple, ext)
-        dist_url = "{}/dists/{}/{}/binary-{}/Packages{}".format(url, dist, comp, arch, ext)
+    for ext, cmd in supported_extensions.items():
+        index = "Packages"
+        index_full = "{}{}".format(index, ext) if ext else index
+
+        output = "{dist}/{comp}/{arch}/{index}".format(
+            dist = dist,
+            comp = comp,
+            arch = arch,
+            index = index,
+        )
+        output_full = "{}{}".format(output, ext) if ext else output
+
+        index_url = "{url}/dists/{dist}/{comp}/binary-{arch}/{index_full}".format(
+            url = url,
+            dist = dist,
+            comp = comp,
+            arch = arch,
+            index_full = index_full,
+        )
+
         download = rctx.download(
-            url = dist_url,
-            output = output,
-            integrity = integrity,
+            url = index_url,
+            output = output_full,
             allow_fail = True,
         )
-        decompress_r = None
-        if download.success:
-            decompress_r = rctx.execute(cmd + [output])
-            if decompress_r.return_code == 0:
-                integrity = download.integrity
-                break
 
-        failed_attempts.append((dist_url, download, decompress_r))
+        if not download.success:
+            reason = "Download failed. See warning above for details."
+            failed_attempts.append((index_url, reason))
+            continue
+
+        if cmd == None:
+            # index is already decompressed
+            break
+
+        decompress_cmd = cmd + [output_full]
+        decompress_res = rctx.execute(decompress_cmd)
+
+        if decompress_res.return_code == 0:
+            break
+
+        reason = "'{cmd}' returned a non-zero exit code: {return_code}"
+        reason += "\n\n{stderr}\n{stdout}"
+        reason = reason.format(
+            cmd = decompress_cmd,
+            return_code = decompress_res.return_code,
+            stderr = decompress_res.stderr,
+            stdout = decompress_res.stdout,
+        )
+
+        failed_attempts.append((index_url, reason))
 
     if len(failed_attempts) == len(supported_extensions):
-        attempt_messages = []
-        for (url, download, decompress) in failed_attempts:
-            reason = "unknown"
-            if not download.success:
-                reason = "Download failed. See warning above for details."
-            elif decompress.return_code != 0:
-                reason = "Decompression failed with non-zero exit code.\n\n{}\n{}".format(decompress.stderr, decompress.stdout)
+        attempt_messages = [
+            "\n  * '{}' FAILED:\n\n  {}".format(url, reason)
+            for url, reason in failed_attempts
+        ]
 
-            attempt_messages.append("""\n*) Failed '{}'\n\n{}""".format(url, reason))
+        fail("Failed to fetch packages index:\n" + "\n".join(attempt_messages))
 
-        fail("""
-** Tried to download {} different package indices and all failed. 
-
-{}
-        """.format(len(failed_attempts), "\n".join(attempt_messages)))
-
-    return ("{}/Packages".format(target_triple), integrity)
+    return rctx.read(output)
 
 def _parse_repository(state, contents, root):
     last_key = ""
@@ -125,12 +148,13 @@ def _create(rctx, sources, archs):
             # on misconfigured HTTP servers)
             url = url.rstrip("/")
 
-            rctx.report_progress("Fetching package index: {}/{} for {}".format(dist, comp, arch))
-            (output, _) = _fetch_package_index(rctx, url, dist, comp, arch, "")
+            index = "{}/{} for {}".format(dist, comp, arch)
 
-            # TODO: this is expensive to perform.
-            rctx.report_progress("Parsing package index: {}/{} for {}".format(dist, comp, arch))
-            _parse_repository(state, rctx.read(output), url)
+            rctx.report_progress("Fetching package index: %s" % index)
+            output = _fetch_package_index(rctx, url, dist, comp, arch)
+
+            rctx.report_progress("Parsing package index: %s" % index)
+            _parse_repository(state, output, url)
 
     return struct(
         package_versions = lambda **kwargs: _package_versions(state, **kwargs),
--- a/apt/private/apt_deb_repository.bzl
+++ b/apt/private/apt_deb_repository.bzl
@@ -124,16 +124,16 @@ def _add_package(state, package):
         vp.append((provides, package))
         util.set_dict(state.virtual_packages, vp, (package["Architecture"], provides["name"]))
 
-def _virtual_packages(state, name, arch):
+def _virtual_packages(state, arch, name):
     return util.get_dict(state.virtual_packages, [arch, name], [])
 
-def _package_versions(state, name, arch):
+def _package_versions(state, arch, name):
     return util.get_dict(state.packages, [arch, name], {}).keys()
 
-def _package(state, name, version, arch):
+def _package(state, arch, name, version):
     return util.get_dict(state.packages, keys = (arch, name, version))
 
-def _create(rctx, sources, archs):
+def _new(rctx, sources, archs):
     state = struct(
         packages = dict(),
         virtual_packages = dict(),
@@ -157,13 +157,13 @@ def _create(rctx, sources, archs):
             _parse_repository(state, output, url)
 
     return struct(
-        package_versions = lambda **kwargs: _package_versions(state, **kwargs),
-        virtual_packages = lambda **kwargs: _virtual_packages(state, **kwargs),
-        package = lambda **kwargs: _package(state, **kwargs),
+        package_versions = lambda arch, name: _package_versions(state, arch, name),
+        virtual_packages = lambda arch, name: _virtual_packages(state, arch, name),
+        package = lambda arch, name, version: _package(state, arch, name, version),
     )
 
 deb_repository = struct(
-    new = _create,
+    new = _new,
 )
 
 # TESTONLY: DO NOT DEPEND ON THIS
@@ -174,9 +174,9 @@ def _create_test_only():
     )
 
     return struct(
-        package_versions = lambda **kwargs: _package_versions(state, **kwargs),
-        virtual_packages = lambda **kwargs: _virtual_packages(state, **kwargs),
-        package = lambda **kwargs: _package(state, **kwargs),
+        package_versions = lambda arch, name: _package_versions(state, arch, name),
+        virtual_packages = lambda arch, name: _virtual_packages(state, arch, name),
+        package = lambda arch, name, version: _package(state, arch, name, version),
         parse_repository = lambda contents: _parse_repository(state, contents, "http://nowhere"),
         packages = state.packages,
         reset = lambda: state.packages.clear(),
--- a/apt/private/apt_deb_repository.bzl
+++ b/apt/private/apt_deb_repository.bzl
@@ -122,7 +122,7 @@ def _add_package(state, package):
 
     # https://www.debian.org/doc/debian-policy/ch-relationships.html#virtual-packages-provides
     if "Provides" in package:
-        provides = version_constraint.parse_dep(package["Provides"])
+        provides = version_constraint.parse_provides(package["Provides"])
 
         state.virtual_packages.add(
             keys = (package["Architecture"], provides["name"]),
--- a/apt/private/apt_deb_repository.bzl
+++ b/apt/private/apt_deb_repository.bzl
@@ -1,6 +1,6 @@
 "https://wiki.debian.org/DebianRepository"
 
-load(":util.bzl", "util")
+load(":nested_dict.bzl", "nested_dict")
 load(":version_constraint.bzl", "version_constraint")
 
 def _fetch_package_index(rctx, url, dist, comp, arch):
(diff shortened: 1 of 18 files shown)


apt/private/package_resolution.bzl
```diff
@@ -1,165 +0,0 @@
-"package resolution"
-
-load(":version.bzl", version_lib = "version")
```
thesayyn, line 3: i don't see how it benefits us to have this file merged into package_index. i'd rather have parsing and resolution done separately.  unless you have a really good reason, i'd say we revert this change.

jjmaestro: I thought the explanation in the commit was a good reason! 😅   Let me copy-paste it here and further explain my train of thought, in case it's helpful.

Previously, in `resolve.bzl` (and IMO the name is already hinting something 😄) we had:
```starlark
    pkgindex = package_index.new(rctx, sources = sources, archs = manifest["archs"])
    pkgresolution = package_resolution.new(index = pkgindex)
    
    constraint = package_resolution.parse_depends(dep_constraint).pop()
```

And now you have:
```starlark
    pkgindex = package_index.new(rctx, sources = sources, archs = manifest["archs"])
    (...)
    constraint = package_index.parse_depends(dep_constraint).pop()
```

`package_resolution` was used only in `resolve.bzl` and only after initializing the `pkgindex`. All of the stuff in `pkgindex` is internal state used to resolve the packages. That's why I think it should all be part of `package_index` which encompasses all of the Debian index functionality.

I think this is nicely shown in the new `_new()`:
```starlark
def _new(rctx, sources, archs):
    index = _index(rctx, sources, archs)

    return struct(
        resolve_all = lambda **kwargs: _resolve_all(index, **kwargs),
        resolve_package = lambda **kwargs: _resolve_package(index, **kwargs),
    )
```

Here, the `index` `struct` is internal and the only thing exposed is "the API", `resolve_all` and `resolve_package`. Which is what we use "outside", because we fetch and parse the index from the manifest and then, we use that state to resolve the packages and populate the lock.

All in all, it's (1) code layout, one file or two files and (2) code architecture / dependency. Before, I think `index` was leaking, because to me it's clearly internal state. IMHO, If we go back to two files, I would still prefer to have the old `package_index` not used in `resolve` and have `package_resolution` use it and only expose the `resolve_*` "API". But then, I think having it all in one file is easier, especially when it's only ~300 lines of code.

Finally, I think that the testing also points at the "internal state" part that I was mentioning. Right now, you pass a mock manifest and then the internals "take care of the index". If you split it, you have to pass the manifest to test `package_index` and then, you would have to either maintain a mock for the `index` `struct` or... you would have to generate it from the mock manifest by calling `package_index.new()` :)

Anyway, let me know what you think!

jjmaestro: In fact, I've just re-checked and I think my point can be made much clearer just looking at the current code! 😅 
```starlark
def _create_resolution(index):
    state = struct(index = index)
    return struct(
        resolve_all = lambda **kwargs: _resolve_all(state, **kwargs),
        resolve_package = lambda **kwargs: _resolve_package(state, **kwargs),
    )
```

It's called state :) `package_resolution` is literally receiving the state from outside instead of receiving `sources` and `archs` and generating it internally, like I do :)

thesayyn: i still don't understand how merging these two together help us as the original thought was that `package_index` is a big state class that holds information in well-structured format for fast access and resolution is a specific algorithm that uses the package_index to perform the resolution using the information from package index.

Currently, you can do this. 

```
index = package_index.new(...)

resolution_a = package_resolution.new(index)

resolution_b = package_resolution.new(index)
```

but once we merge them into one, we automatically lose the ability, and boundaries between the two. 

> It's called state 

We'll it's a big confusing name to use in the case of `package_index`, it's just that way for consistency, it easily could have been index being passed around. 

I am open to renaming some things here and there but i don't see how merging them into one will help us.

thesayyn: In fact, i hope accidental direct commit to main makes things clearer; https://github.com/GoogleContainerTools/rules_distroless/commit/a1b5fe36fa7e1b6b6ffc97e90ccf00f6778fdb72

jjmaestro: > In fact, i hope accidental direct commit to main makes things clearer; [a1b5fe3](https://github.com/GoogleContainerTools/rules_distroless/commit/a1b5fe36fa7e1b6b6ffc97e90ccf00f6778fdb72)

I'm already having quite a big of a rebase with #59... 😭 😭 😭  I'll see if it makes things clearer... 

To my point earlier, I think the (IMHO) mixed mess of:

https://github.com/GoogleContainerTools/rules_distroless/blob/a1b5fe36fa7e1b6b6ffc97e90ccf00f6778fdb72/apt/tests/resolution_test.bzl#L108-L125

and

https://github.com/GoogleContainerTools/rules_distroless/blob/a1b5fe36fa7e1b6b6ffc97e90ccf00f6778fdb72/apt/private/apt_deb_repository.bzl#L142-L160

show exactly what I "predicted" in my comment:

> Finally, I think that the testing also points at the "internal state" part that I was mentioning. Right now, you pass a mock manifest and then the internals "take care of the index". If you split it, you have to pass the manifest to test package_index and then, you would have to either maintain a mock for the index struct or... you would have to generate it from the mock manifest by calling package_index.new() :)

Having to expose internals of (using the old names) `package_index`, mocking them in "prod code" (having the mock in `package_index.bzl` to be able to call `_methods`, etc... to me, this is a pretty big smell.

> i still don't understand how merging these two together help us as the original thought was that package_index is a big state class that holds information in well-structured format for fast access and resolution is a specific algorithm that uses the package_index to perform the resolution using the information from package index.

Precisely! it's an implementation detail, so it's internal and as such, you test it via `resolution`.

Anyway, the rebase with the changes you made in $59 is huge... I'll have to rewrite the PR anyway. So, I'll leave the merging of this stuff for the end or just leave it out completely.

jjmaestro: Finally, e.g. from a very quick look, `version_constraints` also has quite a bit of stuff that should be in `version.bzl` (like I actually did, the `relop` stuff). Plus renaming tests from #59 e.g. `resolution_test.bzl` IMHO should be named like the file it tests, and similarly, the `version_constraint` testing should go to its own test file, which I have also done.

I'll try to get another version of this PR done for review over the week, but meanwhile could we coordinate a bit to avoid breaking the PR? I've been holding for months on this PRs and having such breaks makes me have to re-do not just this PR but the stack... 🙏  

Thanks!

jjmaestro: @thesayyn OK, reading the renaming you did in a1b5fe3 plus

> the original thought was that package_index is a big state class that holds information in well-structured format for fast access and resolution is a specific algorithm that uses the package_index to perform the resolution using the information from package index.

I'll clean up around this and leave them separate, and as mentioned, I'll rename test files, etc, for tests to match the files they test, etc.

jjmaestro:
@thesayyn I've re-done the PR on top of all of the new changes. I've now left the separation of `apt_deb_repository.bzl` and `apt_dep_resolver.bzl`.

I'll wait before rebasing the other PRs just in case you want changes in this one.

Please enable the checks workflow and review it when you can.

Thanks!

jjmaestro:
@thesayyn the changes were done to accommodate the stack of diffs on top, where I've refactored the lockfile (#95), the manifest (#94) the templates to deb_import (#96) and finally the architecture targets (#100) that you just landed in #115.

I'm sorry that you feel many of these are just "because it feels better". It would be good if you could point at the things you feel are that way.

jjmaestro:
> There are some fixes here worth, i'd be happy to land those if you feel like splitting them.

This is probably the easiest, let me know what you want from the stack and I'll start splitting them into PRs.

apt/private/nested_dict.bzl
```diff
@@ -0,0 +1,45 @@
+"nested dict"
+
+def _set(store, keys, value, add = False):
```
thesayyn, line 3: We can revert this, this is more involved than i would like the util .set/get_dict to be.

apt/private/util.bzl
```diff
@@ -1,25 +1,5 @@
 "utilities"
 
-def _set_dict(struct, value = None, keys = []):
```
thesayyn, line 3: same as above, revert

apt/private/version_constraint.bzl
```diff
@@ -57,29 +68,26 @@ def _parse_depends(depends_raw):
...
-    return _version_relop(va[1], vb[1], va[0])
+def _parse_provides(provides_raw):
+    provides = _parse_dep(provides_raw)
+
+    if not provides["version"]:
+        return provides
+
+    op, version = provides["version"]
+
+    if op != "=":
```
thesayyn, line 79: we can't do this, is_satisfied_by checks to see if two different version constraint satisfied where as relop just checks two different versions by an operator.

apt/private/version.bzl
```diff
@@ -114,6 +114,14 @@ def _version_cmp_part(va, vb):
                 return res
     return 0
 
+VERSION_OPERATORS = {
```
thesayyn, line 117: there is no reason for this to be a map, as its not going to grow overtime or change. I'd  rather see this reverted as there is no value here.

apt/private/apt_deb_repository.bzl
```diff
@@ -122,39 +142,24 @@ def _create(rctx, sources, archs):
...
-        package = lambda **kwargs: _package(state, **kwargs),
-        parse_repository = lambda contents: _parse_repository(state, contents, "http://nowhere"),
-        packages = state.packages,
-        reset = lambda: state.packages.clear(),
+        package_versions = lambda arch, name: state.packages.get((arch, name), {}).keys(),
+        virtual_packages = lambda arch, name: state.virtual_packages.get((arch, name), []),
+        package = lambda arch, name, version: state.packages.get((arch, name, version)),
     )
 
-DO_NOT_DEPEND_ON_THIS_TEST_ONLY = struct(
```
thesayyn, line 158: Again, i don't see why this needs to be change, it's already working and there is no value in changing things without a reason. If this change is needed in a larger context, eg if  your stack has a fix for something that needs this change, please put it right next to that fix to help me reason why this needs to change.

apt/private/deb_resolve.bzl
```diff
@@ -64,20 +63,17 @@ def internal_resolve(rctx, yq_toolchain_prefix, manifest, include_transitive):
...
+                arch,
+                constraint["name"],
+                constraint["version"],
                 include_transitive = include_transitive,
             )
 
             if not package:
                 fail("Unable to locate package `%s`" % dep_constraint)
 
-            if len(unmet_dependencies):
```
thesayyn, line 77: I don't see how this a leak, that api is carefully designed this way to not depend much on rctx, now it's being passed as an argument to resolve. 

Previously its up to the caller to do what needs to be done with unmet dependencies but it now it's always warning!


And again if this needs to be change as a prefactor for a fix, please include this with that PR.

apt/tests/apt_dep_resolver_test.bzl
```diff
@@ -0,0 +1,162 @@
...
+
+load("@bazel_skylib//lib:unittest.bzl", "asserts", "unittest")
+load("//apt/private:apt_deb_repository.bzl", "deb_repository")
+load("//apt/private:apt_dep_resolver.bzl", "dependency_resolver")
+load("//apt/tests:apt_deb_repository_test.bzl", idx_mock_setup = "new_setup")
+load("//apt/tests:mocks.bzl", "mock")
+
+_TEST_SUITE_PREFIX = "apt_dep_resolver/"
+
+def _pkg(name, **kwargs):
```
thesayyn, line 11: i like these helper function, we can keep these.

apt/tests/util_test.bzl
```diff
@@ -0,0 +1,26 @@
+"unit tests for util"
```
thesayyn, line 1: i like these new tests.

apt/private/deb_resolve.bzl
jjmaestro: I thought this codebase was the only user of itself, hence why I thought moving the warning to "inside" was equivalent to the current code but "not leaking". I agree on passing `rctx`, I also don't like it much but I thought it was better this way.

I'll revert this then.

apt/private/nested_dict.bzl
jjmaestro: I thought it made sense to have it as a first-class data struct rather than just a helper function on a dict. The struct ensures the access to the dict is through the get-set while just having the functions doesn't prevent any direct change to the dict.

Also, I was using the functionality in other places in the code (e.g. the lockfile refactor) so having it all encapsulated in one data struct made more sense to me.

Do you still prefer it back to naked functions over a dict?

apt/private/version.bzl
jjmaestro: As I mentioned in the commit, the value was in being able to validate the strings against the string keys. But I can remove it back to what it was.

apt/private/version_constraint.bzl
jjmaestro: I just moved the validation to the parsing so that then, we know that all versions in Provides have `=` and thus, we can use the same logic as in relop. Also, doing this makes the code for virtual packages in resolve to be pretty much the same as the real packages.

apt/private/apt_deb_repository.bzl
jjmaestro: This was in the same spirit as the test helpers that I added, standardizing on having a `_TEST_SUITE_PREFIX` constant, etc.

Basically, it was meant to be a clean way to link the stuff that's not publicly accessible for testing purposes only. This way, we always import the module in the test and test the public parts of the struct and if there are parts that are private and we want to test, I thought it would be easier to have a `__test__` sub-struct in-place and not have to create another struct, with a long name in there (the `__test__` already does signal that it's for testing only, I think).

I'll go back to the `DO_NOT_DEPEND_ON_THIS_TEST_ONLY`...

jjmaestro: @thesayyn I'm really looking into this, I was going to make the change but here's the thing. As mentioned, once I've added proper mocking, the tests don't need any of the `_create_test_only()` functionality. That is, all of this functionality is superseded by the mocks, which is the "added value".

So, TL;DR added value == better testing, like with the helper stuff you like in the tests. Is that OK? 🙏

apt/private/version.bzl
jjmaestro: This is the relevant commit I was referring to: 4f504f439e32823d94ec2268f4459c89688948f0

```diff
diff --git a/apt/private/version_constraint.bzl b/apt/private/version_constraint.bzl
index 93e13e5..fe0e6b4 100644
--- a/apt/private/version_constraint.bzl
+++ b/apt/private/version_constraint.bzl
@@ -2,11 +2,22 @@
 
 load(":version.bzl", version_lib = "version")
 
-def _parse_version_constraint(rawv):
-    vconst_i = rawv.find(" ")
-    if vconst_i == -1:
-        fail('invalid version string %s expected a version constraint ">=", "=", ">=", "<<", ">>"' % rawv)
-    return (rawv[:vconst_i], rawv[vconst_i + 1:])
+def _parse_version_constraint(version_and_constraint):
+    chunks = version_and_constraint.split(" ")
+
+    if len(chunks) != 2:
+        fail("Invalid version constraint %s" % version_and_constraint)
+
+    version_constraint = chunks[0]
+    if version_constraint not in version_lib.VERSION_OPERATORS:
+        msg = "Invalid version constraint: %s\nValid constraints are: %s"
+        fail(msg % (version_constraint, version_lib.VERSION_OPERATORS))
+
+    version = chunks[1]
+
+    version_lib.parse(version)  # parsing version to validate it
+
+    return version_constraint, version
 
 def _parse_dep(raw):
     raw = raw.strip()  # remove leading & trailing whitespace
```

Previously, the "validation" was just checking for a space and if not found, it was assuming that the version didn't have a constraint.

With my changes, it (1) checks for the space and fails if none (2) it validates the operator and (3) it validates the version itself.

Is this enough reason for it to be a map? :-?

apt/private/version_constraint.bzl
jjmaestro: I'm going to bring back `_is_satisfied_by`, leaving `_parse_provides` so that we get both improvements (the validation at parsing and the semantics and validation of `_is_satisfied_by` at the time of comparison.

jjmaestro:
I've rebased the branch on top of main, and per your comments, rolled back my changes to `_is_satisfied_by` and the "leaking" of unmet dependencies.

I've added more context in your other reviews re.

* `nested_dict`: TL;DR I also use it in the next PR in the stack so IMO it's better as its own struct VS helper methods that would leave the nested dict "naked" and available to changes via "non-API" (the get-set + other methods offered in the new `nested_dict` struct).

* version op map: I use the keys to validate the operator in the commit I linked. If we go back to a struct of functions I think it would require to add a "validate op" method and other code that IMO is not needed by just using "the best data struct" to organize the operator lookup.

* `__test__` once I added the mocks and helpers that you mentioned you like, the `DO_NOT_DEPEND_ON_THIS_TEST_ONLY` struct becomes dead code.

Again, sorry for such a big PR of mix-and-match stuff, it's definitely hard to review it all, regardless of how much context I added in the commits. Thanks a ton for your reviews, happy to discuss this further and continue making the changes you want.

jjmaestro:
Ran a rebase executing tests and found an issue that I've corrected in the latest push.

jjmaestro:
Just rebased to remove 2eda7c0 (superseded by #129, now merged in `main`)

thesayyn:
@jjmaestro these PRs have stalled for a long time, i don't think i have the mental capacity to review these. 

So i am going to close this and all the stacked ones to clean my plate. If you'd like to send them again, please make one PR at a time with no stacks. 

Thank you for the effort.

PR #95, opened 2024-09-19
refactor: lockfile

jjmaestro:
> [!NOTE]  
>  Stacked on top of #93 (I moved this PR down in the stack)

### refactor: lockfile, new `pkg.bzl`, and lock v1-to-v2 auto-migration
* Refactor lockfile into v2 and add tests. The v2 lockfile format:
  * uses the `nested_dict` `struct` to store the packages so it doesn't need the fast_package_lookup dict.
  * has the dependencies sorted so the lockfile now has stable serialization and the diffs of the lock are actually usable and useful to compare with the changes to the manifest.
  * removes the package and dependency key from the lockfile, now it's done via an external function (`make_deb_import_key` in `deb_import.bzl`)

* Remove `add_package_dependency` from the lockfile API. Now, the package dependencies are passed as an argument to add_package. This way, the lockfile functionality is fully contained in `lockfile.bzl` and e.g. we can remove the "consistency checks" that were only needed because users could forget to add the dependency as a package to the lockfile.

* Ensure backwards-compatibility by internally converting lock v1 to v2. Also, when a lock is set and it's in v1 format, there's a reminder that encourages the users to run @repo//:lock to update the lockfile format.

* Move all of the "package logic" to `pkg.bzl`
* Add tests for `pkg.bzl`
* Add `mock_value` `struct` to `mocks.bzl` to organize the large mock values used for testing `lockfile`, etc.
* Finally, deb_import_key is centralized in deb_import and uses pkg.key
* 
### chore: migrate repo locks to v2
By separating the migration from the previous commit we make sure that the change is backwards compatible:

1. in the previous commit, run all tests with the new code while locks are still v1
2. update the locks n this commit to V2 so we can then re-run all tests in the final state.

### DIFF
--- a/apt/apt.bzl
+++ b/apt/apt.bzl
@@ -126,8 +126,10 @@ def _apt_install(
     )
 
     if not lock and not nolock:
-        # buildifier: disable=print
-        print("\nNo lockfile was given, please run `bazel run @%s//:lock` to create the lockfile." % name)
+        print(
+            "\nNo lockfile was given. To create one please run " +
+            "`bazel run @{}//:lock`".format(name),
+        )
 
     _deb_translate_lock(
         name = name,
--- a/apt/extensions.bzl
+++ b/apt/extensions.bzl
@@ -1,6 +1,6 @@
 "apt extensions"
 
-load("//apt/private:deb_import.bzl", "deb_import")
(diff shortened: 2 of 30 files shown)


jjmaestro:
I've rebased the PR again, no changes from before apart from the rebase.

Issue #56, opened 2024-06-15
Supporting non standard package sources


ericlchen1:
Hey, I ran into some issues when trying to use the nvidia repo as a source url  https://developer.download.nvidia.com/compute/cuda/repos/debian12/x86_64. This is due to the way urls are parsed in rules_distroless and are used to find the Packages file. Also there doesn't seem to be much support when the packages are not multiarch.

Currently I've put together a pretty nonstandard solution with a patch to work around this which supports my use case. However, I was wondering if this would be supported in the future.

rules_distroless.patch
```
diff --git a/apt/private/package_index.bzl b/apt/private/package_index.bzl
index 620a021..c441d4d 100644
--- a/apt/private/package_index.bzl
+++ b/apt/private/package_index.bzl
@@ -2,19 +2,24 @@
 
 load(":util.bzl", "util")
 
-def _fetch_package_index(rctx, url, dist, comp, arch, integrity):
+def _fetch_package_index(rctx, url, dist, comp, arch, integrity, override_url = None):
     target_triple = "{dist}/{comp}/{arch}".format(dist = dist, comp = comp, arch = arch)
     output = "{}/Packages.xz".format(target_triple)
+    url = "{}/dists/{}/{}/binary-{}/Packages.xz".format(url, dist, comp, arch)
+    if override_url:
+        url = override_url
+        output = "{}/Packages".format(target_triple)
     r = rctx.download(
-        url = "{}/dists/{}/{}/binary-{}/Packages.xz".format(url, dist, comp, arch),
+        url = url,
         output = output,
         integrity = integrity,
     )
-    rctx.execute([
-        "xz",
-        "--decompress",
-        output,
-    ])
+    if not override_url:
+        rctx.execute([
+            "xz",
+            "--decompress",
+            output,
+        ])
     return ("{}/Packages".format(target_triple), r.integrity)
 
 def _parse_package_index(state, contents, arch, root):
@@ -59,9 +64,11 @@ def _create(rctx, sources, archs):
     )
 
     for arch in archs:
-        for (url, dist, comp) in sources:
+        for (url, dist, comp, override_url, override_arch) in sources:
+            if override_arch and override_arch != arch:
+                continue
             rctx.report_progress("Fetching package index: {}/{}".format(dist, arch))
-            (output, _) = _fetch_package_index(rctx, url, dist, comp, arch, "")
+            (output, _) = _fetch_package_index(rctx, url, dist, comp, arch, "", override_url)
 
[...]

PR #97, opened 2024-09-19
feat: support flat repos

jjmaestro:
> [!NOTE]  
>  Stacked on top of #96

Fixes issue #56 and #66

Follow-up and credit to @alexconrey (PR #55), @ericlchen1 (PR #64) and @benmccown (PR #67) for their work on similar PRs that I've reviewed and drawn some inspiration to create "one 💍 PR to merge them all" 😅

## Problem

Debian has two types of repos: "canonical" and "flat". Each has a different sources.list syntax:

"canonical": (see https://wiki.debian.org/DebianRepository/Format#Overview)
```
deb uri distribution [component1] [component2] [...]
```

flat: (see https://wiki.debian.org/DebianRepository/Format#Flat_Repository_Format)
```
deb uri directory/
```

Per the spec,
> A flat repository does not use the dists hierarchy of directories, and instead places meta index and indices directly into the archive root (or some part below it)

Thus, [the URL logic in `_fetch_package_index()`](https://github.com/GoogleContainerTools/rules_distroless/blob/2ce7b477def75579c49bab25266f953f30275c88/apt/private/package_index.bzl#L21) is incorrect for these repos and it always fails to fetch the Package index.

## Solution

Just use the Debian sources.list convention in the `sources` section of the manifest to add canonical and flat repos. Depending on whether the channel has one directory that ends in `'/'` or a `(dist, component, ...)` structure the `_fetch_package_index ()` and other internal logic will know whether the source is a canonical or a flat repo.

For example:
```yaml
version: 1

sources:
  # canonical repo
  - channel: bullseye main contrib
    url: https://snapshot-cloudflare.debian.org/archive/debian/20240210T223313Z
  # flat repos, note the trailing '/' and the lack of distribution or components
  - channel: bullseye-cran40/
    url: https://cloud.r-project.org/bin/linux/debian
  - channel: ubuntu2404/x86_64/
    url: https://developer.download.nvidia.com/compute/cuda/repos

archs:
  - amd64

packages:
  - bash
  - r-mathlib
  - nvidia-container-toolkit-base
```

Disregarding the "mixing" of Ubuntu and Debian repos for the purpose of the example, this manifest shows that you can mix canonical and flat repos and you can mix multiarch and single-arch flat repos and canonical repos.

You will still have the same problems as before with packages that only exist for one architecture and/or repos that only support one architecture. In those cases, simply separate the repos and packages into their own manifests.

----------

> [!NOTE]
> This PR also fixes an issue with NVIDIA CUDA flat repos that don't follow the Debian repo spec and have invalid `'Filename'` paths.

[The Debian repo spec for `'Filename'`](https://wiki.debian.org/DebianRepository/Format#Filename) says:
> The mandatory Filename field shall list the path of the package archive relative to the base directory of the repository. The path should be in canonical form, that is, without any components denoting the current or parent directory ("." or ".."). It also should not make use of any protocol-specific components, such as URL-encoded parameters.

However, there are cases where this is not honored. In those cases we try to work around this by assuming `'Filename'` is relative to the `sources.list` `directory/` so we combine them and normalize the new `'Filename'` path.

Note that, so far, only the [NVIDIA CUDA repos](https://developer.download.nvidia.com/compute/cuda/repos/) needed this workaround so maybe this heuristic will break for other repos that don't conform to the Debian repo spec.


jjmaestro:
See previous discussions in #86

thesayyn:
I don't think we can land this without a proper api first. So i am going to close this to clean up here. https://github.com/GoogleContainerTools/rules_distroless/issues/123

Issue #98, opened 2024-09-19
bug: packages that already have data.tar.gz break


jjmaestro:
```
ERROR: Traceback (most recent call last):
        File "/home/jjmaestro/.cache/bazel/_bazel_jjmaestro/e0806ab47903450e6ee36289894189ae/external/cuda_nsight-systems-2022.4.2_amd64_2022.4.2.50-32196742v0/BUILD.bazel", line 4, column 8, in <toplevel>
                genrule(                                                                                    
Error in genrule: rule 'data' has file 'data.tar.gz' as both an input and an output  
ERROR: /home/jjmaestro/.cache/bazel/_bazel_jjmaestro/e0806ab47903450e6ee36289894189ae/external/cuda_nsight-systems-2022.4.2_amd64_2022.4.2.50-32196742v0/BUILD.bazel: no such target '@@cuda_nsight-systems-2022.4.2_amd
64_2022.4.2.50-32196742v0//:data': target 'data' not declared in package '' defined by /home/jjmaestro/.cache/bazel/_bazel_jjmaestro/e0806ab47903450e6ee36289894189ae/external/cuda_nsight-systems-2022.4.2_amd64_2022.4
.2.50-32196742v0/BUILD.bazel                                                                                                                                                                                            
ERROR: /home/jjmaestro/.cache/bazel/_bazel_jjmaestro/e0806ab47903450e6ee36289894189ae/external/cuda/nsight-systems-2022.4.2/amd64/BUILD.bazel:3:6: no such target '@@cuda_nsight-systems-2022.4.2_amd64_2022.4.2.50-3219
6742v0//:data': target 'data' not declared in package '' defined by /home/jjmaestro/.cache/bazel/_bazel_jjmaestro/e0806ab47903450e6ee36289894189ae/external/cuda_nsight-systems-2022.4.2_amd64_2022.4.2.50-32196742v0/BU
ILD.bazel and referenced by '@@cuda//nsight-systems-2022.4.2/amd64:data'                                    
ERROR: Analysis of target '//examples/cuda:test' failed; build aborted: Analysis failed
INFO: Elapsed time: 0.075s, Critical Path: 0.00s   
INFO: 1 process: 1 internal.                                                                                
ERROR: Build did NOT complete successfully                                                                  
ERROR: No test targets were found, yet testing was requested
```

[...]

PR #99, opened 2024-09-19
fix: Debian packages with tar.gz data file

jjmaestro:
When a Debian package has a data file that's already compressed with `gzip` and named `data.tar.gz` it fails with:
```
Error in genrule: rule 'data' has file 'data.tar.gz' as both an input and an output.
```

See issue #98

### DIFF
--- a/apt/private/deb_import.bzl
+++ b/apt/private/deb_import.bzl
@@ -3,14 +3,45 @@
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
 # BUILD.bazel template
-_DEB_IMPORT_BUILD_TMPL = """
-# TODO: https://github.com/bazel-contrib/rules_oci/pull/523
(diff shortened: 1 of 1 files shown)


jjmaestro:
It passes all the current tests and I've manually tested it in a development branch with one of the problematic CUDA packages. This fix seems to work and the CUDA packages don't fail like they did before. However, I can't get that test to pass as it fails on yet-another issue that I'm trying to debug.

Still, I think the diff is easy to follow and I think it's good to go.

jjmaestro:
@thesayyn I think this fix should go first and then I'll rebase the refactor and other branches. What do you think?

thesayyn (approved):

thesayyn:
Bear with me as i work through these PRs :)  FYI It may take days/weeks.

jjmaestro:
@thesayyn of course!! no worries :)

PR #100, opened 2024-09-20
feat: platforms-aware package arch aliases

jjmaestro:
> [!NOTE]  
>  Stacked on top of #97 

# feat: platforms-aware package arch aliases
Package repos can now be used directly without the `/<ARCH>` bit.

The "root-level" `BUILD` in the package repo now has platforms-aware aliases that use `select` to pick the right arch for the `:data`, `:control`, etc., targets.

Now, you can do:

```starlark
load("@rules_oci//oci:defs.bzl", "oci_image")

oci_image(
    name = "foo",
    base = "@distroless_base_nossl_debian12",
    tars = [
        "@debian12//coreutils",
        "@debian12//bash",
    ],
)
```

when before you would have needed something like:

```starlark
load("@rules_oci//oci:defs.bzl", "oci_image")

CPU_ARCH = [
    ("arm64", "arm64"),
    ("x86_64", "amd64"),
]

PACKAGES = [
    "@debian12//coreutils",
    "@debian12//bash",
]

PACKAGES_ARCH = select({
    "@platforms//cpu:%s" % cpu: [
        "%s/%s" % (package, arch)
        for package in PACKAGES
    ]
    for cpu, arch in CPU_ARCH
})

oci_image(
    name = "foo",
    base = "@distroless_base_nossl_debian12",
    tars = PACKAGES_ARCH,
)
```

To force a specific architecture the "arch packages" can still be used just like before with `@repo//package/arch` or via constraints and platforms (`--platforms=...`).

### chore: remove `select()` from the examples
This is a separate diff so that we can test the previous diff with the changes plus the examples with `select` in place and make sure everything works as-is with the new package aliases.


jjmaestro:
Closing, superseded by #115 and #121.

PR #105, opened 2024-10-23
dpkg_status should use fixed mtime

alexeagle:
Fixes #106


jrockway:
Great minds think alike.

thesayyn (approved):

PR #107, opened 2024-10-24
test modified at timestamps

sallustfire:
@alexeagle I added some tests to the debian and ubuntu examples, if you want them. Unfortunately, container structure test doesn't let you check the modified at timestamps, so it requires a heavier command test.


examples/debian_snapshot/test_linux_amd64.yaml
```diff
@@ -19,10 +19,22 @@ commandTests:
       - perl/now 5\.32\.1-4\+deb11u3 amd64 \[installed,local\]
       - tzdata/now 2024a-0\+deb11u1 all \[installed,local\]
       - nvidia-kernel-common/now 20151021\+13 amd64 \[installed,local\]
+  - name: "stat dpkg status"
```
jjmaestro, line 22: Please `cp examples/debian_snapshot/test_linux* e2e/smoke/` so the `e2e` tests are in-sync with `debian_snapshot`. Thanks!

thesayyn:
I have no idea why the CI is not running for this. Probably because only diff is on yaml files?

PR #115, opened 2024-11-20
feat: add convience targets

thesayyn:
Currently targets generated by apt is really crude, forces all packages to be enumerated again in the BUILD file. 

This PR improves the API surface by introducing a few new convenience targets.

- `@bullseye//:bullseye` target at the external repository root which has the same name as the repo that contains all the `installed packages` + a `dpkg_status target` for all installed packages to be found by scanners.
-  `@bullseye//bash` target that is alias to architecture specific packages with a `select` to allow same oci_image target to be transitioned to different platforms.
-  `@bullseye//:packages` target: to access all package archives (including transitives)
-  `@bullseye//:dpkg_status` target: to access a dpkg_status archive (including transitives)


thesayyn:
cc @jjmaestro

jjmaestro:
@thesayyn so, part of this is a big chunk of "the top of the stack" of PRs (#100)... and again, given the amount of changes that this will introduce, I will have to re-do a ton of the intermediate PRs in the stack that have been 2 months waiting to be reviewed.

Just FYI, it's very demotivating to see a big chunk of the functionality that made me revamp the repo being implemented without the rest of the changes. And again, this IMHO "slaps it in" while I built the refactor to clean all of the parts. Again, "2 months in the making".

jjmaestro:
Just more context of how I would have done something like this... at the very least, you could have eyeballed the stack and considered if you are (mostly) going to merge it in or not. And if so, you would have implemented this on top of #100, adding the extra convenience aliases that you are adding.

That would help a contributor like me to feel like the work that's been done is worth something. Instead, I now have to scramble and rebase everything with larger conflicts.

thesayyn:
I didn't see you made a similar improvement. apologies on my end for that.  Big PRs are hard to review and i don't have the time to look at them at once. In my defense refactor PRs are mostly "i don't like the way you did things so i am gonna just change it".  I am not trying to be extra, if you look at [this ](https://github.com/GoogleContainerTools/rules_distroless/pulls?q=is%3Apr+is%3Aclosed)  you will see that i took the time to actually review and land a lot of PRs.

And the reason pinged you was to get your opinions as you were trying to make this ruleset better. :)

thesayyn:
It's also fair to mention that i am not getting paid or being thanked either for the most of the work i do here.

jjmaestro:
IMO refactoring is definitely not "I don't like how you did this". Most of the time it should be to make the code more maintainable so that other changes become easier. That's how I've tried to make all the changes in the stack.

jjmaestro:
> It's also fair to mention that i am not getting paid or being thanked either for the most of the work i do here. 

I know, neither are any of the contributors! But that's also why I go above and beyond and document the PRs, what they do, each commit has a (hopefully) clear message and concrete change, etc. So that, regardless of how large the PR / stack, it's easy to see the "conducting thread" and have a breadcrumb / trail of everything.

And that's why it's disconcerting to see that after 2 months, the reviews (which are good) don't look at the whole stack. 

Anyway, enough drama. For my part, sorry for dumping this on you. 

Will go over the changes over this week and will try to get the PRs in the stack rebased.

apt/private/deb_translate_lock.bzl
```diff
@@ -20,12 +22,87 @@ _DEB_IMPORT_TMPL = '''\
...
+_ARCHITECTURE_MAP = {{
+    "amd64": "x86_64",
+    "arm64": "aarch64",
+    "ppc64el": "ppc",
+    "mips64el": "mips64"
+}}
+
+_ARCHITECTURES = {architectures}
+
+# See officially supported platforms at https://www.debian.org/releases/stable/armel/ch02s01.en.html
```
jjmaestro, line 55: From #100:
https://github.com/jjmaestro/rules_distroless/blob/b0f06d770cb6702179aa7fc994d1d1541cb4e8f7/apt/private/util.bzl#L3-L15

I added it in `util.bzl` instead of putting it inside the repo. Any advantages to doing this? I guess it's so you can put the `config_setting` in the `BUILD`. I still don't know enough to understand the advantages but it looks good here. I'd change the comment and the map, though. I think mine is a better fit.

apt/private/deb_translate_lock.bzl
```diff
@@ -42,13 +119,20 @@ def _deb_translate_lock_impl(rctx):
         if len(lockf.packages()) < 1:
             package_defs.append("   pass")
 
+    # TODO: rework lockfile to include architecure information
+    architectures = sets.make()
+    packages = sets.make()
```
jjmaestro, line 124: This re-work is what I did in #95 and later added in #100 the `get_architecture` for all this part, see:

* https://github.com/jjmaestro/rules_distroless/blob/b0f06d770cb6702179aa7fc994d1d1541cb4e8f7/apt/private/deb_translate_lock.bzl#L36
* https://github.com/jjmaestro/rules_distroless/blob/b0f06d770cb6702179aa7fc994d1d1541cb4e8f7/apt/private/lockfile.bzl#L47-L55

apt/private/deb_translate_lock.bzl
```diff
@@ -78,8 +162,29 @@ def _deb_translate_lock_impl(rctx):
             ),
         )
 
+    # TODO: rework lockfile to include architecure information and merge these two loops
```
jjmaestro, line 165: That's exactly what I did in https://github.com/jjmaestro/rules_distroless/blob/b0f06d770cb6702179aa7fc994d1d1541cb4e8f7/apt/private/deb_translate_lock.bzl#L25-L40. Also, a big part of the cleanup in this file comes from another refactor, #96.

apt/private/starlark_codegen_utils.bzl
```diff
@@ -0,0 +1,45 @@
+"Utilities for generating starlark source code"
```
jjmaestro, line 1: This is pretty cool, I like it being extracted out. I have all of this in [`deb_import.bzl`](https://github.com/jjmaestro/rules_distroless/blob/b0f06d770cb6702179aa7fc994d1d1541cb4e8f7/apt/private/deb_import.bzl) (coming from #96 and #100).

It's all inside the helpers and it makes the code look complicated (the output is nicely formatted Starlark as well :) but it can probably be done much better with this approach.

jjmaestro (commented):
Reviewing the PR, and acknowledging I'm very biased because of the outstanding PRs I have done... it definitely looks like this could be implemented easily / easier and IMO nicer on top of #100  😕  It's basically #100 plus some aliases that are quite useful and a very nice approach to generate the Starlark code (VS my crappier ad-hoc approach).

It's probably much to ask but I would appreciate if we could re-do this on top of #100, taking all the cleanups for the `lockfile` and `deb_import`. I really like the `starlark_codegen_utils.bzl` part, I think it would clean up a lot of the code I wrote in `deb_import` to do the templates.

I'll probably try to re-implement this on top of #100 to see how it looks like, I would really like to try and have it all on the PR stack to avoid more churn (it's taken me days to re-do the PRs 😞 )

examples/debian_snapshot/BUILD.bazel
```diff
@@ -54,66 +54,52 @@ cacerts(
...
-)
-
 oci_image(
-    name = "apt",
-    architecture = select({
-        "@platforms//cpu:arm64": "arm64",
-        "@platforms//cpu:x86_64": "amd64",
-    }),
+    name = "image",
+    architecture = "arm64",
```
jjmaestro, line 59: Shouldn't this be reverted so that `architecture` is still a `select` so that we build and test two architectures?

Or is that done via the `platform_transition` and this is just a placeholder? :-?

examples/debian_snapshot/BUILD.bazel
```diff
@@ -54,66 +54,52 @@ cacerts(
...
     env = {
         # Required to use the SSL certs from `cacerts()`
         "SSL_CERT_FILE": "/etc/ssl/certs/ca-certificates.crt",
     },
     os = "linux",
     tars = [
+        # This target contains all the installed packages.
+        "@bullseye//:bullseye",
```
jjmaestro, line 67: just a nit, but I guess the packages should go at the end of the `tars` so that if they change, the previous layers remain the same.

jjmaestro:
Also, the tests are failing, here's the patch to fix it:

```diff
diff --git a/apt/private/BUILD.bazel b/apt/private/BUILD.bazel
index 19a4af1..cadebf9 100644
--- a/apt/private/BUILD.bazel
+++ b/apt/private/BUILD.bazel
@@ -27,6 +27,8 @@ bzl_library(
     visibility = ["//apt:__subpackages__"],
     deps = [
         ":lockfile",
+        ":starlark_codegen_utils",
+        "@bazel_skylib//lib:new_sets",
         "@bazel_tools//tools/build_defs/repo:cache.bzl",
         "@bazel_tools//tools/build_defs/repo:http.bzl",
         "@bazel_tools//tools/build_defs/repo:utils.bzl",
@@ -95,3 +97,9 @@ bzl_library(
     visibility = ["//apt:__subpackages__"],
     deps = [":version"],
 )
+
+bzl_library(
+    name = "starlark_codegen_utils",
+    srcs = ["starlark_codegen_utils.bzl"],
+    visibility = ["//apt:__subpackages__"],
+)
```

jjmaestro:
Also found a bug that breaks in `amd64`, here's the fix:
```diff
diff --git a/apt/private/deb_translate_lock.bzl b/apt/private/deb_translate_lock.bzl
index 6491868..509062c 100644
--- a/apt/private/deb_translate_lock.bzl
+++ b/apt/private/deb_translate_lock.bzl
@@ -169,11 +169,11 @@ def _deb_translate_lock_impl(rctx):
             _PACKAGE_TEMPLATE.format(
                 target_name = package["name"],
                 data_targets = starlark_codegen_utils.to_dict_attr({
-                    "//:linux_%s" % arch: "//%s/%s" % (package["name"], package["arch"])
+                    "//:linux_%s" % arch: "//%s/%s" % (package["name"], arch)
                     for arch in architectures._values
                 }),
                 control_targets = starlark_codegen_utils.to_dict_attr({
-                    "//:linux_%s" % arch: "//%s/%s:control" % (package["name"], package["arch"])
+                    "//:linux_%s" % arch: "//%s/%s:control" % (package["name"], arch)
                     for arch in architectures._values
                 }),
             ),
```

apt/private/deb_translate_lock.bzl
thesayyn: It's almost same except that its not in two places and easy to follow.

apt/private/deb_translate_lock.bzl
thesayyn: Yeah, i'd be happy land that, but for now, i'll just workaround this.

examples/debian_snapshot/BUILD.bazel
thesayyn: Yes, this is just metadata, platform_transition_filegroup transitions to arm64 always.

jjmaestro: I've checked and it seems that this is not correct. The resulting image has the architecture always set to `arm64` even when the image is built for `amd64` :-/

PR #118, opened 2024-11-27
fix: only test against 7.x and 8.x

thesayyn:


meteorcloudy:
"8.x" doesn't work yet since 8.0.0 is not released yet. We just implemented "N.*" with bazelisk, which will also fetch rc.

thesayyn:
> "8.x" doesn't work yet since 8.0.0 is not released yet. We just implemented "N.*" with bazelisk, which will also fetch rc.

Thank you.

PR #120, opened 2024-12-02
fix: platforms, architectures and dpkg_status after #115

jjmaestro:
After landing #115 we need `platforms` to be non-dev. This would have been caught in the e2e test but since #115 didn't make  changes to the e2e test, it went unnoticed.

To prevent this, I've also copied the `examples/debian_snapshot` to the `e2e` test. IMHO this helps catch bugs that affect third-party repos using `rules_distroless` as a dependency, such as the `platforms` dependency.

I've also verified that we do need the `select` in the `architecture`, as mentioned in my [code review](https://github.com/GoogleContainerTools/rules_distroless/pull/115#discussion_r1855644693). Otherwise, the `Config` in the manifest is always set to `arm64` regardless of the architecture that we are building.

I've also fixed the unmapped Debian architectures (e.g. "all"). Basically, `_ARCHITECTURE_MAP[arch] or arch` fails when a Debian arch is not in the map. It should be `_ARCHITECTURE_MAP.get(arch, arch)` to default to `arch` when it's not in the map. To exercise this and ensure it always works I've:

* removed "arm64" since it's a valid CPU platform already (and without this fix it would have already failed in all the examples and the e2e test).
* added `armhf` mapping to `armv7e-mf`
* changed Debian's `ppc64el` mapping to the exact matching platforms CPU (`ppc64le`)
* added a test to exercise arch `all` (which is both an arch that's not in the map while being "a special Debian arch" and also a valid platforms CPU constraint).
* Moved architecture doc links next to `_ARCHITECTURE_MAP` and added a bunch more links to the Debian wikis.

Finally, I've fixed `_PACKAGES`, `:packages` and `:dpkg_status` to be different per-architecture, because there are cases where the packages differ between architectures. E.g. currently in the `examples/ubuntu_snapshot`, `coreutils` depends on `libssl3` on `amd64` but there's no such dependency on `arm64`. Thus, the list of packages and `dpkg_status` has to be different.

PR #121, opened 2024-12-02
feat: add more convenience targets

jjmaestro:
Add platform-aware `:data` and `:control` convenience targets at the top of the package repo so that we can use `@<REPO>//<PACKAGE>:data` directly. See changes to `examples/debian_snapshot/BUILD`:

```diff
diff --git a/examples/debian_snapshot/BUILD.bazel b/examples/debian_snapshot/BUILD.bazel
index 88b98ba..298c261 100644
--- a/examples/debian_snapshot/BUILD.bazel
+++ b/examples/debian_snapshot/BUILD.bazel
@@ -58,10 +58,7 @@ tar(
 
 cacerts(
     name = "cacerts",
-    package = select({
-        "@platforms//cpu:arm64": "@bullseye//ca-certificates/arm64:data",
-        "@platforms//cpu:x86_64": "@bullseye//ca-certificates/amd64:data",
-    }),
+    package = "@bullseye//ca-certificates:data",
 )
 
 oci_image(
```

Also, add `:deps` to complement `:data`.

Finally, with the changes in #115 and the previous commit, we can now remove a bunch of select from the examples and use the new target aliases.

NOTE: This PR adds to #115 some functionality that I had in #100 and we can close #100 completely.


apt/private/deb_translate_lock.bzl
```diff
@@ -23,43 +23,37 @@ _DEB_IMPORT_TMPL = '''\
...
+#     * https://www.debian.org/releases/stable/amd64/ch02s01.en.html#idm186
+#
+# For more info on Bazel's platforms CPUs see:
+#     * https://github.com/bazelbuild/platforms/blob/main/cpu/BUILD
 _ARCHITECTURE_MAP = {{
     "amd64": "x86_64",
-    "arm64": "aarch64",
-    "ppc64el": "ppc",
-    "mips64el": "mips64"
+    "armhf": "armv7e-mf",
```
thesayyn, line 44: Can you send this as a separate PR?

apt/private/deb_translate_lock.bzl
```diff
@@ -107,6 +101,76 @@ filegroup(
 )
 """
 
+def _package_build(template, package, repo_name):
```
thesayyn, line 104: Okay as part of this PR this doesn't has to change and it makes it 10x harder to understand a PR that could have been really simple to understand.  Can you revert this so i can actually see the diff?

jjmaestro:
@thesayyn sorry for the noise :( I'll back out the rebase and pull the last fix that changes the architecture map into a separate PR.

apt/private/deb_translate_lock.bzl
jjmaestro: Ah, this is already in #120, it came in with the rebase :(

jjmaestro:
@thesayyn the noise comes from the stacked PRs, when I rebased this PR on top of #120, I didn't realize it would make it much harder to review 😞

I'm used to stacked PRs and rebases "just working" and GH is pretty terrible when it comes to this, I didn't realize it would dump all of the contents from your PR in here as changes :-/

Can you review #120 first? Otherwise, you can focus on just this two commits::
 * b953eb7 - feat: add more convenience targets
 * 5c38a79 - chore: remove select() from the examples

jjmaestro:
OK, I managed to back out the rebase, this PR is now "clean" and separate from #120.

I'm really really sorry for all the noise, I'm still getting used to working with GH's limitations :(

apt/private/deb_translate_lock.bzl
```diff
@@ -107,6 +93,76 @@ filegroup(
...
```
thesayyn, line 96: can you revert this change? I'd like to see the actual diff rather than the refactor.

jjmaestro: Ah! I misunderstood your comments, I thought the issue was with the rest of the code from the rebase that made it hard to review!

I'll make another proposal.

jjmaestro:
@thesayyn let me know if the PR is to your liking, I'll be happy to make more changes if needed. Thanks.

jjmaestro:
@thesayyn I've further simplified the PR removing all of the "refactored" methods, so now it's just clear that the PR ditched the `_PACKAGE_TEMPLATE` in favor of reusing `package_template`.

Let me know if the PR is to your liking, I'll be happy to make more changes if needed. Thanks.

apt/private/deb_translate_lock.bzl
```diff
@@ -23,20 +23,6 @@ _DEB_IMPORT_TMPL = '''\
     )
 '''
 
-_PACKAGE_TEMPLATE = """\
```
thesayyn, line 26: Let's keep this as private, your change makes it reuse the public template, which will break usages such as; 

https://github.com/GoogleContainerTools/distroless/blob/main/private/repos/deb/package.BUILD.tmpl

Let's keep the PR focused, it says "add more convience targets", so it should be doing just that.

apt/private/deb_translate_lock.bzl
```diff
@@ -169,12 +188,16 @@ def _deb_translate_lock_impl(rctx):
             "%s/BUILD.bazel" % (package["name"]),
             _PACKAGE_TEMPLATE.format(
                 target_name = package["name"],
-                data_targets = starlark_codegen_utils.to_dict_attr({
-                    "//:linux_%s" % arch: "//%s/%s" % (package["name"], arch)
+                data_targets = "select(%s)" % starlark_codegen_utils.to_dict_attr({
```
thesayyn, line 191: can you put back the select statement into the template?

apt/private/deb_translate_lock.bzl
```diff
@@ -169,12 +188,16 @@ def _deb_translate_lock_impl(rctx):
...
+                    for arch in architectures._values
+                }),
+                control_targets = "select(%s)" % starlark_codegen_utils.to_dict_attr({
+                    "//:linux_%s" % arch: "//%s/%s:control" % (package["name"], package["arch"])
                     for arch in architectures._values
                 }),
-                control_targets = starlark_codegen_utils.to_dict_attr({
-                    "//:linux_%s" % arch: "//%s/%s:control" % (package["name"], arch)
+                deps = "select(%s)" % starlark_codegen_utils.to_dict_list_attr({
+                    "//:linux_%s" % arch: ["//%s/%s:deps" % (package["name"], package["arch"])]
```
thesayyn, line 200:
```suggestion
                    "//:linux_%s" % arch: ["//%s/%s:deps" % (package["name"], package["arch"])]
```

since `:deps` target exclude the package, doesn't this create a broken transitive closure? i am not sure what's the point of `deps` here.

jjmaestro: The point is to have the package dependencies separate and available, similar to how `:data` is the actual package itself. Then, you have the full set of `:deps` and `:data` available as the `target_name`.

thesayyn: Yes, but what i am pointing out is that, `:data` is current package, and `:deps` should be everything the transitive closure of the current package, but it's not the case here, because `:deps` doesn't have the transitive closure of the transitives. So it's missing a turtle when traversing down the closure.  

Or am i misunderstanding this new `:deps` target?

jjmaestro: `:deps` is exactly the same as what was before in `{deps}`, right? Just "one level up" so that it `select`s the right arch and available as a separate target label.

jjmaestro: Let me see if this clarifies it more:

```diff
diff --git a/apt/private/package.BUILD.tmpl b/apt/private/package.BUILD.tmpl
index 9c3dd61..8810a96 100644
--- a/apt/private/package.BUILD.tmpl
+++ b/apt/private/package.BUILD.tmpl
@@ -1,17 +1,28 @@
(...)
+
+filegroup(
+    name = "deps",
+    srcs = {deps},
     visibility = ["//visibility:public"],
 )
 
 filegroup(
     name = "{target_name}",
+    srcs = [
+        ":deps",
+        ":data",
+    ],
     visibility = ["//visibility:public"],
-    srcs = {deps} + [":data"]
 )
 ```
 
 That's the change in `package.BUILD.tmpl` and in `_PACKAGE_TEMPLATE`.
 
 Since I haven't changed how `deps` is generated, I don't see how it can be missing anything in the transitive closure :-?
 
 As mentioned, `:deps` is a convenience target that is literally what was before in `{deps}`. Now, also in `_PACKAGE_TEMPLATE` which is "one level up" so `@<REPO>//<PACKAGE>:deps` will point to `@<REPO>//<PACKAGE>/<ARCH>:deps` through a `select`.
 
 Let me know if this makes sense. Thanks!!

thesayyn: i understand the change, i still don't understand how's that useful and in what context. deps is literally just all data dependencies of a package, an example might make justify adding it. otherwise it's something we have to support forever.

jjmaestro: OK, I'll remove it and go back to `rcs = {deps} + [":data"]` then!

jjmaestro: Ah, I just pushed the change and it breaks... because it's actually "needed", the alias at the top of the repo call the per-arch `:deps` in the "inner" per-arch repos...

I'm going to try to fix it removing `:deps` but I really think it should be added... all that it does is just expose the same stuff that's in the `Dependencies` part of the package. I don't see how "supporting this forever" would be hard :-?

jjmaestro: OK, I just pointed the `:package_name` target in the upper part of the package repo to the `:arch` target in the "inner" per-arch repos, so no need to have `:deps`.

I still think it would be nice to have, but this way is much simpler and you prefer it so here we go!

jjmaestro:
@thesayyn I'm going to rebase the branch on `main` to remove the "Conflicting files" message from the threads but you now have the "clean Compares" to check.

Learning to work around GH limitations... 😅

jjmaestro:
@thesayyn and, again, after I did the two pushes before the rebase, somehow GH is not showing the "Compare" for the two of them, only for the last one!?? This makes it impossible to see the changes I did to address your comments... this is absurd, I really don't know how GH can be failing at this stuff o.0.

Here's my "manual compare link" between the changes: [`Compare 3327243..b198392`](https://github.com/GoogleContainerTools/rules_distroless/compare/3327243..b198392).

thesayyn (approved):
Looking good, lets rebase and land.

jjmaestro:
Rebased again and fixed errors in the code that were failing the tests, it should be green now!

PR #131, opened 2024-12-09
feat: update flatten to deduplicate files

westrik:
#119 added a deduplicate option to `flatten` to address "duplicates of file paths not supported" errors when loading flattened images with Docker.

This option deduplicates directories, but in its current implementation it does not deduplicate files. This means the duplicate file path errors can still occur in some cases.

This change updates `flatten.sh` to deduplicate files as well, with the final file in the archive being preserved for each path.

It also renames `assert_tar_listing` to `assert_tar_mtree` and creates a new `assert_tar_listing` rule that lists archive contents instead of exporting an mtree.

### DIFF
--- a/distroless/private/flatten.sh
+++ b/distroless/private/flatten.sh
@@ -36,8 +36,8 @@ if [[ "$output" != "-" ]]; then
     $bsdtar --confirmation "$@" > $output 2< <(awk '{
         count[$1]++;
         files[NR] = $1
-     }
-     END {
+    }
+    END {
         ORS=""
         for (i=1; i<=NR; i++) {
             seen[files[i]]++
--- a/distroless/private/flatten.sh
+++ b/distroless/private/flatten.sh
@@ -17,7 +17,6 @@ if [[ "$output" != "-" ]]; then
         fi
     done
 
-    
     # There not a lot happening here but there is still too many implicit knowledge.
     # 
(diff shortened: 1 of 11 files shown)


distroless/private/flatten.sh
```diff
@@ -29,21 +28,26 @@ if [[ "$output" != "-" ]]; then
     # See: https://github.com/libarchive/libarchive/blob/f745a848d7a81758cd9fcd49d7fd45caeebe1c3d/tar/util.c#L240
     # See: https://github.com/libarchive/libarchive/blob/f745a848d7a81758cd9fcd49d7fd45caeebe1c3d/tar/util.c#L216
     # 
-    # And finally we iterate over all the entries generating 31 bytes of interleaved 'Y' or 'N' date based on if 
-    # we came across the entry before, for directories the first occurrence is kept, and for files copies are 
-    # preserved.
+    # To match the extraction behavior of tar itself, we want to preserve only the final occurrence of each file
```
thesayyn, line 31: i am not sure if keeping the last occurrence the directory is the right thing to do, overlayfs is probably smart enough to do the right thing here though. I can also see this being right thing to do as later entries might have different mode, time etc.

thesayyn: This is probably fine now that i think about it.

thesayyn (approved):

PR #134, opened 2024-12-11
feat: support multiple URLs in sources

### LINKED ISSUE #113: FR: Support alternate URLs for package_index
We are using `apt` in [apt.bzl](https://github.com/GoogleContainerTools/rules_distroless/blob/main/apt/apt.bzl) to create a Debian package index. Our CI went down because the snapshot repository we were using, https://snapshot-cloudflare.debian.org, went down. https://snapshot.debian.org was still up - it would be nice to have the ability to add it as a mirror URL. I think it would require updating `_fetch_package_index` in [package_index.bzl](https://github.com/GoogleContainerTools/rules_distroless/blob/main/apt/private/package_index.bzl) to support multiple URLs.

jjmaestro:
Fixes #113

The Debian snapshot mirror fails quite often and, while the Cloudflare mirror has had some issues in the past (e.g. it was lagging replication for months) it's much more reliable.

This commit adds support for multiple URLs in the manifest. See examples/debian_snapshot/bullseye.yaml for a sample usage.

### DIFF
--- a/apt/private/lockfile.bzl
+++ b/apt/private/lockfile.bzl
@@ -20,7 +20,10 @@ def _add_package(lock, package, arch):
         "key": k,
         "name": package["Package"],
         "version": package["Version"],
-        "url": "%s/%s" % (package["Root"], package["Filename"]),
+        "urls": [
+            "%s/%s" % (root, package["Filename"])
+            for root in package["Roots"]
+        ],
         "sha256": package["SHA256"],
         "arch": arch,
         "dependencies": [],
@@ -71,6 +74,10 @@ def _from_json(rctx, content):
         fast_package_lookup = dict(),
     )
     for (i, package) in enumerate(lock.packages):
+        # TODO: only support urls before 1.0
+        if "url" in package:
+            package["urls"] = [package.pop("url")]
+
         lock.packages[i] = package
         lock.fast_package_lookup[package["key"]] = i
     return _create(rctx, lock)
--- a/e2e/smoke/bullseye.lock.json
+++ b/e2e/smoke/bullseye.lock.json
@@ -6,7 +6,10 @@
 			"key": "ncurses-base_6.2-p-20201114-2-p-deb11u2_amd64",
 			"name": "ncurses-base",
 			"sha256": "a55a5f94299448279da6a6c2031a9816dc768cd300668ff82ecfc6480bbfc83d",
-			"url": "https://snapshot-cloudflare.debian.org/archive/debian/20240210T223313Z/pool/main/n/ncurses/ncurses-base_6.2+20201114-2+deb11u2_all.deb",
+			"urls": [
+				"https://snapshot-cloudflare.debian.org/archive/debian/20240210T223313Z/pool/main/n/ncurses/ncurses-base_6.2+20201114-2+deb11u2_all.deb",
+				"https://snapshot.debian.org/archive/debian/20240210T223313Z/pool/main/n/ncurses/ncurses-base_6.2+20201114-2+deb11u2_all.deb"
+			],
 			"version": "6.2+20201114-2+deb11u2"
 		},
 		{
@@ -41,7 +44,10 @@
 			"key": "libncurses6_6.2-p-20201114-2-p-deb11u2_amd64",
 			"name": "libncurses6",
 			"sha256": "5b75c540d26d0525f231d39e5cf27ea7919d57305ba7101ea430c975369095eb",
-			"url": "https://snapshot-cloudflare.debian.org/archive/debian/20240210T223313Z/pool/main/n/ncurses/libncurses6_6.2+20201114-2+deb11u2_amd64.deb",
+			"urls": [
+				"https://snapshot-cloudflare.debian.org/archive/debian/20240210T223313Z/pool/main/n/ncurses/libncurses6_6.2+20201114-2+deb11u2_amd64.deb",
+				"https://snapshot.debian.org/archive/debian/20240210T223313Z/pool/main/n/ncurses/libncurses6_6.2+20201114-2+deb11u2_amd64.deb"
+			],
 			"version": "6.2+20201114-2+deb11u2"
 		},
 		{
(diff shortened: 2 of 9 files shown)


thesayyn:
You need a rebase here; so far looks okay to me.

e2e/smoke/bullseye.lock.json
```diff
@@ -6,7 +6,7 @@
 			"key": "ncurses-base_6.2-p-20201114-2-p-deb11u2_amd64",
 			"name": "ncurses-base",
 			"sha256": "a55a5f94299448279da6a6c2031a9816dc768cd300668ff82ecfc6480bbfc83d",
-			"url": "https://snapshot.debian.org/archive/debian/20240210T223313Z/pool/main/n/ncurses/ncurses-base_6.2+20201114-2+deb11u2_all.deb",
+			"url": "https://snapshot-cloudflare.debian.org/archive/debian/20240210T223313Z/pool/main/n/ncurses/ncurses-base_6.2+20201114-2+deb11u2_all.deb",
```
thesayyn, line 9: I would expect the lock file to have a `urls` field as well, since urls specified in the manifest is interchangeable and bazel will fallback to next one if one of these links fail. 

So we need a breaking change to the lockfile here.

jjmaestro: I went with the single URL because I think it's the better source of truth. Meaning, while it's true that the intent of the manifest is to point at equivalent packages, the lock is *the actual stuff downloaded*, and so the `sha256` corresponds to *that exact deb file downloaded* from *that exact URL*. Thus, IMO is actually correct as-is and I think adding multiple URLs wouldn't be right.

WDYT?

jjmaestro: OK, I'll make the change and will also look into fixing the buildifier errors

jjmaestro:
This commit fixes the buildifier issues. I'll make the changes to the lock now.

jjmaestro:
@thesayyn thinking more about the lockfile, I was considering the "breaking change" and wondering if you would want / prefer the approach I did in #95: supporting `v1` and a new `v2`, having auto-migration, etc.

If so, maybe you could consider (1) landing this PR as-is and (2) either checking the last comments I did on #93 so I could move that forward and then add the `urls` to #95... or I could also re-do #95 from scratch without depending on #93.

I'm just thinking that it would be nice to avoid the breaking change... let me know what you think!

thesayyn:
Since we are pre 1.0 i'd be okay with just breaking things, that said maybe we could be less sensitive about `url` vs `urls` in the lockfile, accept both?

jjmaestro:
> maybe we could be less sensitive about url vs urls in the lockfile, accept both?

true, true, sounds like a good compromise for now, I think! Same way the manifest can have both... 

OK, I'll go about this and we can go over the other PRs later on (ah, that huge stack... I would definitely do things differently now XXD)

jjmaestro:
@thesayyn I've added `urls` to the lock, I've tested it manually with the original lockfiles and with an updated lockfile. Also, I've left all the lockfiles in "the old format" (with `url`), let me know if you want to update them all now or leave them as-is for later.

e2e/smoke/bullseye.yaml
```diff
@@ -11,12 +11,13 @@ version: 1
...
   - channel: bullseye main contrib
-    url: https://snapshot-cloudflare.debian.org/archive/debian/20240210T223313Z
+    urls:
+      - https://snapshot-cloudflare.debian.org/archive/debian/20240210T223313Z
+      - https://snapshot.debian.org/archive/debian/20240210T223313Z
   - channel: bullseye-security main
     url: https://snapshot-cloudflare.debian.org/archive/debian-security/20240210T223313Z
   - channel: bullseye-updates main
     url: https://snapshot-cloudflare.debian.org/archive/debian/20240210T223313Z/
-  # This channel contains an uncompressed Packages file
```
thesayyn, line 19: Hmm, why remove?

examples/debian_snapshot/bullseye.yaml
```diff
@@ -11,7 +11,9 @@ version: 1
 
 sources:
   - channel: bullseye main contrib
-    url: https://snapshot-cloudflare.debian.org/archive/debian/20240210T223313Z
+    urls:
```
thesayyn, line 14: Do you want to run lock for this?

e2e/smoke/bullseye.lock.json
thesayyn: Its determined at download time.  Lets say you have to urls where an artifact can be fetched, bazel will try both of those urls sequentially until it succeeds, if it can't download it via any of the urls, then it will fail the build. 

But if you determine this a locking time and generate a lockfile with a single location where a package might be downloaded, then Bazel will fail immediately if the said url no longer works.

thesayyn: Think of urls array as a list of locations where a package might be fetched, any of them could go offline after locking.

jjmaestro: Yeah, I mean, I understand and I'm aware of `urls` as a concept and how it's meant to work 😄  I was just considering the lock as being "the actual thing that happened"...  basically, I was considering a "what if" there's a difference in hash between the URLs, but that's also covered, we are using the hash in the lock and passing it to verify the download, in https://github.com/GoogleContainerTools/rules_distroless/blob/2354eb3345e7f332abb0d2ff43bd8399fabca731/apt/extensions.bzl#L39

So all is good 😄 and yeah, having all the possible URLs in the lock is better and more correct.

e2e/smoke/bullseye.yaml
jjmaestro: To keep `e2e/smoke/` in-sync with `examples/debian_snapshot`. Remember I tried the symlink but you thought it would possibly break the BCR... it would be great to have a way of doing this check at commit time so that they are always kept in-sync, but for now, when I have changes in any of these, I sync them by hand.

examples/debian_snapshot/bullseye.yaml
jjmaestro: I pushed the commit with the "old locks" so you could run the CI with them being old & the new code, and then the CI with both new code and new locks :)  But it's OK, I did test both by hand locally, plus I think the code is clear enough that both cases do work.

apt/private/lockfile.bzl
```diff
@@ -71,6 +74,9 @@ def _from_json(rctx, content):
         fast_package_lookup = dict(),
     )
     for (i, package) in enumerate(lock.packages):
+        if "url" in package:
```
thesayyn, line 77: leave TODO: for us to remove before 1.0?

e2e/smoke/bullseye.yaml
thesayyn: Hmm, i see, we might want to separate these at some point so that no sync is necessary..

thesayyn (approved):
LGTM, one small nit

e2e/smoke/bullseye.yaml
jjmaestro: It would be great to just run all of the tests in the project and avoid duplicates. Although I guess there's probably some "Bazel reason(s)" as to why this setup is the way it is!

PR #135, opened 2024-12-11
fix: pre-commit run config and broken README

jjmaestro:
### chore: update .pre-commit-config.yaml
    
Steps:
* `pre-commit migrate-config`
* `pre-commit autoupdate`
    
Then moving prettier from `v4.0.0-alpha.8` to the latest stable version in the mirror, `v3.1.0`, because the alpha versions seem to produce a node_modules directory somehow.

Note that the prettier mirror repo is archived, it seems unsupported / unmaintained since Dec 2023 :-/

### fix: README.md

The pre-commit run --all-files that I did and rebased into #120 broke the README :(

dadc64c added some of the fixes and this commit adds the CAUTION warning plus a TIP warning to highlight the need of small and readable PRs.

Now, prettier runs without breaking this Markdown.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,7 +4,7 @@
 
 # Commitizen runs in commit-msg stage
 # but we don't want to run the other hooks on commit messages
-default_stages: [commit]
+default_stages: [pre-commit]
 
 # Use a slightly older version of node by default
 # as the default uses a very new version of GLIBC
@@ -14,18 +14,18 @@ default_language_version:
 repos:
   # Check formatting and lint for starlark code
   - repo: https://github.com/keith/pre-commit-buildifier
-    rev: 6.1.0.1
+    rev: 7.3.1.1
     hooks:
       - id: buildifier
       - id: buildifier-lint
   # Enforce that commit messages allow for later changelog generation
   - repo: https://github.com/commitizen-tools/commitizen
-    rev: v2.18.0
+    rev: v4.1.0
     hooks:
       # Requires that commitizen is already installed
       - id: commitizen
         stages: [commit-msg]
   - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: "v2.4.0"
+    rev: v3.1.0
     hooks:
       - id: prettier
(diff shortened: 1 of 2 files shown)


thesayyn:
No need to fix these, as it doesn't seem to make the CI upset.

jjmaestro:
But... the pre-commit hook actually complaints 😅  That is, these fixes come from `pre-commit run --all-files` so... they should be committed to the repo, right? :-?

I don't know why they weren't caught before though, as I mentioned in #120 I was suspecting my setup, but dunno.

If not landing these then I guess the `.pre-commit-config` should be changed? I was also going to update it since I got 3e6477e landed in [keith/pre-commit-buildifier](https://github.com/keith/pre-commit-buildifier/)...

jjmaestro:
Looking at the changes, it's true that the "fixes" are... weird o.0  it actually "breaks" parts of the markdown :S

I'll dig a bit more into this and will comment here with findings, in case you want to change the pre-commit config.

jjmaestro:
And since I rebased #120 those "pre-commit fixes" went in... 😢 I'm going to "fix the fixes" to the README and fix the pre-commit config to exclude `README` from `prettifier` because it seems to not be able to process the GH Markdown correctly or something :-?

jjmaestro:
@thesayyn I've updated this PR to fix what I broke and to fix pre-commit config and the README so that this stuff hopefully won't break in the future.

Sorry for the noise... I hope that at least updating the config is worth it 😞

jjmaestro:
@thesayyn I see you fixed parts of the README yesterday in dadc64c :) I've added the CAUTION and another TIP to highlight the short PR, plus now pre-commit runs without breaking the README.

jjmaestro:
> No need to fix these, as it doesn't seem to make the CI upset.

Agreed, but when running `pre-commit run --all-files` it was making changes and broke the Markdown in the README.

I think these should go in, I'm fixing the README putting back the right casing in the links, adds back the note highlights and add the TIP highlight to stress the need  of small PRs, and I upgraded the pre-commit hooks to (hopefully) improve `prettier` linting (and e.g. bring in 3e6477e from [keith/pre-commit-buildifier](https://github.com/keith/pre-commit-buildifier/)).

Let me know what you think!

P.S. even now, when you run pre-commit manually on `main`, it's linting the README and fixing one whitespace change 😅

thesayyn (approved):

PR #139, opened 2024-12-11
fix: //docs tests in MacOS

jjmaestro:
Fixes #138

PR #148, opened 2025-01-22
fix: #121

jjmaestro:
In #121 I introduced new aliases and when using one of them (`:data`) in `cacerts`, we are hitting the new `select`s that only have Linux branches and no default case, so it breaks in non-linux OSes.

The `cacerts` rule needs to be restricted to only compatible OSes.

### DIFF
--- a/e2e/smoke/BUILD
+++ b/e2e/smoke/BUILD
@@ -12,7 +12,14 @@ load("@aspect_bazel_lib//lib:tar.bzl", "tar")
 load("@aspect_bazel_lib//lib:transitions.bzl", "platform_transition_filegroup")
 load("@container_structure_test//:defs.bzl", "container_structure_test")
 load("@rules_distroless//distroless:defs.bzl", "cacerts", "group", "passwd")
-load("@rules_oci//oci:defs.bzl", "oci_image")
+load("@rules_oci//oci:defs.bzl", "oci_image", "oci_load")
+
+COMPATIBLE_WITH = select({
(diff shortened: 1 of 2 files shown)


jjmaestro:
@thesayyn I just noticed the error when I rebased and was trying to repro / check #139 running the tests in OSX. I thought CI was running checks for MacOS too, is that not the case? :-?

jjmaestro:
If I'm reading this correctly, the CI is only excluding Windows: https://github.com/GoogleContainerTools/rules_distroless/blob/2354eb3345e7f332abb0d2ff43bd8399fabca731/.github/workflows/ci.yaml#L27-L31

But looking at [the last CI run in #121](https://github.com/GoogleContainerTools/rules_distroless/actions/runs/12917341172/job/36023819889) it seems like the matrix step only picked Ubuntu and skipped Mac and Windows... maybe it's because the runner only scheduled it in a linux server? :-?

Dunno, GH Actions are a bit hard to debug... in any case, this fixes the issue with the tests, but it would be good to find out if the CI is running everywhere it should!

jjmaestro:
I see [the CI ran on MacOS](https://github.com/GoogleContainerTools/rules_distroless/actions/runs/12917967651) but only after the merge! Is that how it's configured? :-?

thesayyn:
> I see [the CI ran on MacOS](https://github.com/GoogleContainerTools/rules_distroless/actions/runs/12917967651) but only after the merge! Is that how it's configured? :-?

it only runs on main, and branch names that contain `mac` since mac runners are expensive.

thesayyn (approved):

PR #149, opened 2025-01-24
fix: mark apt extension as reproducible when using a lockfile

cerisier:
When using a lockfile, an entry in the MODULE.bazel.lock is redundant.

Marking the extension as reproducible only when a lockfile is provided saves an entry in the lockfile.

Note, I had to move `bazel_features` out of `dev_dependency` because the `reproducible` field on `extension_metadata` only appeared in `7.1.0`.


cerisier:
@thesayyn buildifier fixed

cerisier:
@thesayyn doc tests fixed

thesayyn:
you have some buildifier errors

cerisier:
Fixed again, sorry about that.

PR #155, opened 2025-02-24
fix: circular dependencies

### LINKED ISSUE #128: FR: Gracefully handle self-edges and circular dependencies.
Currently, we don't check for circular dependencies and self-edges. For instance `libc6` package depends on `libcrypt1` which in turn depends on `libc6` which is circular and not permitted in Bazel.

birunts:
Prevent adding root package as a dependency to itself.

Resolves #128 & Closes #128.

### DIFF
--- a/apt/private/apt_dep_resolver.bzl
+++ b/apt/private/apt_dep_resolver.bzl
@@ -122,7 +122,8 @@ def _resolve_all(state, name, version, arch, include_transitive = True):
         if key in already_recursed:
             continue
 
-        if i != 0:
+        # Do not add dependency if it's a root package to avoid circular dependency.
+        if i != 0 and key != root_package["Package"]:
             # Add it to the dependencies
             already_recursed[key] = True
             dependencies.append(package)
--- a/apt/tests/resolution_test.bzl
+++ b/apt/tests/resolution_test.bzl
@@ -255,10 +255,73 @@ def _resolve_aliases(ctx):
 
 resolve_aliases_test = unittest.make(_resolve_aliases)
 
+def _resolve_circular_deps_test(ctx):
+    env = unittest.begin(ctx)
+
+    idx = _make_index()
+
+    # `ruby` dependencies should have no `ruby`
+    idx.add_package(package = "libruby")
+    idx.add_package(package = "ruby3.1", depends = "ruby")
+    idx.add_package(package = "ruby-rubygems", depends = "ruby3.1")
+    idx.add_package(package = "ruby", depends = "libruby, ruby-rubygems")
+
+    (root_package, dependencies, _) = idx.resolution.resolve_all(
+        name = "ruby",
+        version = "",
+        arch = _test_arch,
+    )
+    asserts.equals(env, "ruby", root_package["Package"])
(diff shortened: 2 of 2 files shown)


apt/private/apt_dep_resolver.bzl
```diff
@@ -122,7 +122,7 @@ def _resolve_all(state, name, version, arch, include_transitive = True):
         if key in already_recursed:
             continue
 
-        if i != 0:
+        if i != 0 and key != root_package["Package"]:
```
thesayyn, line 125: do you mind leaving a comment here that explains whats happening?

apt/tests/resolution_test.bzl
```diff
@@ -255,10 +255,36 @@ def _resolve_aliases(ctx):
...
+resolve_circular_deps_test = unittest.make(_resolve_circular_deps_test)
+
 _TEST_SUITE_PREFIX = "package_resolution/"
 
 def resolution_tests():
     parse_depends_test(name = _TEST_SUITE_PREFIX + "parse_depends")
     resolve_optionals_test(name = _TEST_SUITE_PREFIX + "resolve_optionals")
     resolve_architecture_specific_packages_test(name = _TEST_SUITE_PREFIX + "resolve_architectures_specific")
     resolve_aliases_test(name = _TEST_SUITE_PREFIX + "resolve_aliases")
+    resolve_circular_deps_test(name = _TEST_SUITE_PREFIX + "parse_circular")
```
thesayyn, line 290: this technically a self-edge case, circular dependency is where a dependency somewhere endsup depending on itself.

self-edge: `libcurl -> libcurl`

circular: `libcurl -> netty -> curl -> libcurl`

birunts: That is what should be within the test:
`ruby -> ruby3.1 -> ruby` 

But now I've extended it to be:
`ruby -> ruby-rubygems -> ruby3.1 -> ruby`

apt/private/apt_dep_resolver.bzl
birunts: Sure :) Done.

apt/tests/resolution_test.bzl
thesayyn: the test might pass but the generated BUILD files will still be circular afaik.

birunts: > but the generated BUILD files will still be circular afaik.

What exactly? Sorry I do not know the `rules_distroless` details, yet :wink: 

If it comes to generated [filegroups](https://github.com/GoogleContainerTools/rules_distroless/blob/main/apt/private/package.BUILD.tmpl#L17) (with this fix) the target won't contain itself in `srcs` attribute anymore.

birunts: Kindly reminder :slightly_smiling_face: @thesayyn any chance to have that reviewed/merged, please?

thesayyn (approved):

thesayyn:
Turns out this did not fix the circular deps case; https://github.com/GoogleContainerTools/rules_distroless/issues/163

PR #164, opened 2025-05-28
fix: filegroup circular deps

### LINKED ISSUE #163: Cycle error when using all-in-one target
Bazel complains about a cycle in the dependency graph when I try to use the all-in-one target and provide it to a `oci_image`

I've created a repro here: https://github.com/purkhusid/distroless_cycle_repro

You can reproduce the cycle error by running `bazel build @bookworm//:bookworm`

birunts:
Previously generated `filegroup` targets can contain dependency cycle.

The case here is that defined filegroup of a package should not have deps of other packages but rather to its `:data` package.

This change removes potential dependency cycles and makes references to package's `:data`.

Resolves #163 & Closes #163.

### DIFF
--- a/apt/private/deb_translate_lock.bzl
+++ b/apt/private/deb_translate_lock.bzl
@@ -39,7 +39,7 @@ alias(
 
 filegroup(
     name = "{target_name}",
-    srcs = select({deps}) + [":data"],
+    srcs = select({deps}),
     visibility = ["//visibility:public"],
 )
 '''
--- a/apt/private/deb_translate_lock.bzl
+++ b/apt/private/deb_translate_lock.bzl
@@ -192,7 +192,7 @@ def _deb_translate_lock_impl(rctx):
                 control_targets = '"@%s//:control"' % repo_name,
                 src = '"@%s//:data"' % repo_name,
                 deps = starlark_codegen_utils.to_list_attr([
-                    "//%s/%s" % (dep["name"], package["arch"])
+                    "//%s/%s:data" % (dep["name"], package["arch"])
                     for dep in package["dependencies"]
                 ]),
                 urls = package["urls"],


thesayyn:
I don't think this solves the problem. By not depending on the `:data` target only, you are excluding the transitive dependencies, that's why you don't see the problem anymore but this basically breaks everything if i am not mistaken.

birunts:
Agree. But currently package `filegroup`s are defined wrong and you cannot even _use_ it (`bazelisk build @bookworm//libgcc-s1:libgcc-s1`). If package should be done this way to contain transitive dependencies, definitely it cannot be `filegroup` but rather custom rule who will be able to handle that somehow. 

From another side gathering all needed [_PACKAGES](https://github.com/GoogleContainerTools/rules_distroless/blob/main/apt/private/deb_translate_lock.bzl#L97) is pointless as if packages should handle transitive dependencies then this list should contain _only_ those explicitly specified in `bookworm.yaml` file where transitive deps will be resolved.
So even better that _real_ filegroup package (e.g. `@bookworm//libgcc-s1/amd64:amd64`) do not contain deps but only its `:data` and 

Indeed _current_ change is not a fix but rather a quick workaround. What I see now, the _real_ package (e.g. `@bookworm//libgcc-s1/amd64:amd64`) is a custom rule who contains package `:data` only and returns a provider with its (Depends,Recommends,Suggests,Breaks) dependencies similarly as dep packages do. 
Then _main_ [package container](https://github.com/GoogleContainerTools/rules_distroless/blob/main/apt/private/deb_translate_lock.bzl#L120) is also a custom rule who can use that provider(s) and produce a target with full list of packages `:data` to be installed. 

At least this is mine understanding of current state :man_shrugging:

birunts:
One more look at the issue :wink: 

I have checked generated package defs and first commit removes redundant `+ [:data]`. It points to the same _target_ as the `:data` within `<package>/<arch>` definition. Actually no big deal. 

When it comes to the second commit all the package's data dependencies should be listed with `<package>:data`. We do not loose any transitive dependencies because the inserted list comes from the lockfile package dependency list, i.e. [package["dependencies"]](https://github.com/bazel-contrib/rules_distroless/blob/v0.6.1/apt/private/deb_translate_lock.bzl#L196), where it covers already all transitive deps of the package. 

Bazel dependency graph is of [Directed Acyclic Graph](https://bazel.build/concepts/dependencies), thus current filegroup's definition is wrong, making cycle in dependency graph. 

@thesayyn & @alexeagle plase have a look once again for the fix.

thesayyn (approved):

PR #167, opened 2025-06-19
Add support for netrc-based auth

gergondet-woven:
Hello and thank you for making this available.

This PR adds support for netrc-based authentication when using Debian repository that might require it.

Prior to this change, the only way to authenticate with such a mirror would be (afaik) to input the credentials in the manifest which would be less than ideal.


RomanVlasenko:
Can't you use `--credential_helper` to supply credentials in this case?
https://bazel.build/reference/command-line-reference#common_options-flag--credential_helper

thesayyn (approved):

PR #178, opened 2025-08-21
fix: relocate tar toolchain

alexeagle:
Newer bazel-lib versions re-export the tar.bzl toolchain type. Since rules_distroless dererefences it from the `tar_lib` helper, we'd need to dep on it directly.

bazel-lib 3 will remove the redirect; to be forwards-compatible we shouldn't rely on it.

Also make our bzl_library targets correct, working around https://github.com/bazelbuild/bazel-skylib/issues/568


alexeagle:
Having trouble greening this up. https://github.com/bazel-contrib/bazel-lib/pull/1153 should be the quicker fix (and principled)

This one should still stay open since we'll want it for eventual bazel-lib 3.0

PR #180, opened 2025-08-21
fix: remove realpath call from decompression logic

thesayyn:
Remove realpath call for data.tar.xz, data.tar.zst, and data.tar.lzma handling.


loosebazooka (approved):

alexeagle (approved):
Thanks, I was going to track that down last night when the CI is so horribly spammy

PR #186, opened 2025-09-18
disable verbose flag to tar

benjaminp:
This was generating spammy Bazel output.


thesayyn:
This was a silly debugging line i forgot in there.

PR #201, opened 2025-11-06
Add riscv64 as an package architecture option

loosebazooka:


loosebazooka:
Hrmm isn't as straightforward as I thought

thesayyn:
What are you trying to do? I can help

loosebazooka:
trying to solve for
```
ERROR: Traceback (most recent call last):
        File "/home/appu/.cache/bazel/_bazel_appu/dd657112d407274a19bd1a60f81d99eb/external/rules_distroless~~apt~trixie/BUILD.bazel", line 42, column 46, in <toplevel>
                "@platforms//cpu:" + _ARCHITECTURE_MAP[arch],
Error: key "riscv64" not found in dictionary
```
which I think is just adding riscv64 to https://github.com/bazel-contrib/rules_distroless/blob/main/apt/private/deb_translate_lock.bzl#L64

riscv is a supported architecture since trixie, and I have users asking for riscv images in distroless (hence the error I'm running into)

loosebazooka:
oof this was just a formatting error?

MODULE.bazel
```diff
@@ -22,6 +22,7 @@ use_repo(yq_toolchains, "yq_linux_amd64")
 use_repo(yq_toolchains, "yq_linux_arm64")
 use_repo(yq_toolchains, "yq_linux_ppc64le")
 use_repo(yq_toolchains, "yq_linux_s390x")
+use_repo(yq_toolchains, "yq_linux_riscv64")
```
thesayyn: i think you don't even need this change, since it makes the ruleset itself work on riscv.

thesayyn (approved):

thesayyn:
> oof this was just a formatting error?

appears to be yes

MODULE.bazel
loosebazooka: I imagine maybe someone would want that? if they were to append things onto the riscv images? Oh like they're not expected to have riscv build machine?

loosebazooka: I guess we can add it back in later. I'll remove it for now.

PR #208, opened 2026-01-30
fix: Allow multiple cc_imports per pkgconfig file

blorente:
Some packages, like `krb5-multidev`, include several .so files to link against in their pkg-config files:
```
prefix=/usr
exec_prefix=${prefix}
libdir=${prefix}/lib/x86_64-linux-gnu/mit-krb5
includedir=${prefix}/include/mit-krb5

defccname=FILE:/tmp/krb5cc_%{uid}
defktname=FILE:/etc/krb5.keytab
defcktname=FILE:/etc/krb5/user/%{euid}/client.keytab

Name: mit-krb5
Description: An implementation of Kerberos network authentication
Version: 1.20.1
Cflags: -isystem ${includedir}
Libs: -L${libdir} -lkrb5 -lk5crypto -lcom_err
Libs.private: -lkrb5support
```

So, instead of assuming that we have one cc_library per pkgconfig file, we just collect all the declarations and create one `cc_import` target for each.

Some design considerations:

### Does this violate the private-ness of `Libs.private`?

All `cc_imports` we create have private visibility already, so I don't think it does.

### Why not depend on other debian packages exporting those libraries?

Since those files are already in the package (https://packages.debian.org/sid/amd64/krb5-multidev/filelist), I assume the intention is to link against the `.so` files distributed in the `krb-multidev` package, and not to pull other deb packages. This could be false, of course, if a user has already installed those packages and for some reason the linker finds their SO files first, but I think that's a bug in the definition of the krb5-multidev package, not this implementation.

Plus, it has the added benefit that, if the debian package in question doesn't include an appropriate .so file, we just don't create the cc_import.

### DIFF
--- a/apt/private/deb_import.bzl
+++ b/apt/private/deb_import.bzl
@@ -272,36 +272,40 @@ so_library(
             includes += pkgc.includes
             link_paths += pkgc.link_paths
 
-            if not pkgc.libname or pkgc.libname + "_import" in import_targets:
+            if len(pkgc.libnames) == 0:
                 continue
 
-            subtarget = pkgc.libname + "_import"
-            import_targets.append(subtarget)
+            for libname in pkgc.libnames:
+              if libname + "_import" in import_targets:
+                continue
 
-            # Look for a static archive
-            # for ar in a_files:
-            #     if ar.endswith(pkgc.libname + ".a"):
-            #         static_lib = '":%s"' % ar
-            #         break
+              subtarget = libname + "_import"
+              import_targets.append(subtarget)
 
-            # Look for a dynamic library
-            IGNORE = ["libfl"]
-            for so_lib in so_files:
-                if pkgc.libname and pkgc.libname not in IGNORE and so_lib.endswith(pkgc.libname + ".so"):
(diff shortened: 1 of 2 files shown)


apt/private/deb_import.bzl
```diff
@@ -272,36 +272,42 @@ so_library(
             includes += pkgc.includes
             link_paths += pkgc.link_paths
 
-            if not pkgc.libname or pkgc.libname + "_import" in import_targets:
+            # TODO BL: not sure what the second part of this condition means
+            # if not pkgc.libname or pkgc.libname + "_import" in import_targets:
```
thesayyn, line 276: as i remember; some packages exported multiple pkgconfig with name libname, (for whatever reason), so this was here to prevent duplicate targets.

thesayyn (approved):

apt/private/deb_import.bzl
blorente: Ohh, okay, I swear pkgconfig reminds me of https://xkcd.com/2347/ :) I've moved that logic below anyway, inside the for loop :)

PR #209, opened 2026-02-16
apt: add hermetic sysroot support for rust sys crate

alok-oai:
Motivation:
Rust `-sys` crates need hermetic apt-provided headers/libs. Some crates can disable build scripts and link directly to Bazel C/C++ targets (for example `libudev-sys`, `alsa-sys`, `gstreamer-*-sys`, `wayland-sys`). Others still require build scripts and explicit sysroot-backed include/library search paths (notably `libspa-sys` / `pipewire-sys`, and `openssl-sys`). GStreamer test/runtime also needs plugin discovery paths.

Changes:
- add and export `apt_sysroot` to materialize apt tar inputs as a sysroot tree
- generate `:sysroot` targets in translated dependency-set outputs
- refactor `deb_import` to inspect `.deb` payloads and improve symlink/library resolution for cross-package linkage
- improve link path handling for pkg-config and multiarch lib directories
- switch apt shared-library linking to DT_RPATH behavior for hermetic transitive resolution
- improve Contents index probing/parsing for Debian/Ubuntu layout differences
- fix Provides-version constraint comparison direction


loosebazooka:
my apologies for my earlier comment, I misread the situation

PR #214, opened 2026-03-24
Normalize paths in flatten

loosebazooka:
This normalizes all paths before flattening.

When flatting tars with equivalent paths `./etc/duck` vs `etc/duck/` flatten would just let them all in the flattened result. loading those images into a local container runtime would result in an error with repeated paths error. This instead normalizes them to `etc/duck` to prevent this.

This doesn't normalize absolute paths like `/etc/duck`

### DIFF
--- a/distroless/private/flatten.sh
+++ b/distroless/private/flatten.sh
@@ -35,16 +35,25 @@ if [[ "$deduplicate" == "True" ]]; then
     # number of occurrences of each path, and the second pass determines whether each entry is the final (or only)
     # occurrence of that path.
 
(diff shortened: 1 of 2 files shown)


thesayyn:
Wouldn't it be better to normalize them to `./`? Its much more readable.

loosebazooka:
I guess I have no preference. I can switch it

thesayyn (approved):

loosebazooka:
@thesayyn I have a followup for this that will just preserve the seen path while deduplicating

PR #216, opened 2026-04-15
fix: normalize paths to ./

thesayyn:


thesayyn:
@codex review

kormide (approved):

dzbarsky (approved):

examples/flatten/BUILD.bazel
loosebazooka, line 115: shouldn't these have now changed?

distroless/private/flatten.sh
```diff
@@ -37,9 +37,11 @@ if [[ "$deduplicate" == "True" ]]; then
 
     $bsdtar --confirmation "$@" 2< <("${awk}" '
```
loosebazooka, line 38: if bsdtar itself doesn't modify the result, then it has no effect on normalization.

thesayyn: ahh right, this will preserve the original path, so i was completely wrong.

loosebazooka: hang on, I have a pr that was just sitting on my computer

PR #221, opened 2026-06-03
Module extension API with fixes and sysroot unpacking 

### LINKED ISSUE #124: FR: SYSROOT generated by rules_distroless for hermetic CC complation
While working on a GLIBC related problem, i realized that rules_distroless could have made hermetic cc compilation so much easier by generating a sysroot out of the debian packages that are being fetched. 

This is great for one single reason, what's actually going into the container is also what you are linking against!

I can imagine a bzlmod only API, as such

```starlark
# Generate a sysroot using the apt extensions
apt = use_extension("@rules_distroless//apt:extensions.bzl", "apt")

apt.install("libc6-dev")

apt.sysroot(
    name = "sysroot_amd64"
    arch = "amd64"
)
 
apt.sysroot(
    name = "sysroot_arm64"
    arch = "arm64"
)
 
use_repo(apt, "sysroot_amd64", "sysroot_arm64")
 
# Use with toolchains_llvm
llvm = use_extension("@toolchains_llvm//toolchain/extensions:llvm.bzl", "llvm")
llvm.sysroot(
    name = "llvm_toolchain_with_sysroot",
    targets = ["linux-x86_64"],
    label = "@sysroot_amd64//:sysroot",
)
llvm.sysroot(
    name = "llvm_toolchain_with_sysroot",
    targets = ["linux-arm64"],
    label = "@sysroot_arm64//:sysroot",
)
```

LittleHuba:
This changeset enables rules_distroless to create sysroots for toolchains_llvm. (Closes #124)

It is based on the previous work by @thesayyn in https://github.com/bazel-contrib/rules_distroless/pull/183.

The work can be split in three distinct pieces:
1. Fixes to the upstream PR (all up to 235e0596e766522469e689438198682447aa564f)
2. Removing gawk (f5f280aa27ce9ee7abc3cf8eda502ded8e5e6d1b and fead4d117f127142ca7f3b034d9fb937df36f69b)
3. Implementing sysroot generation (remaining commits)

## Fixes to upstream

I first focused on getting the repository into a stable state by fixing infrastructure to make as many tests succeed as possible.

This is followed up by fixes to the implementation of the new API.
Main issue was mixing of architectures of the different packages that lead to conflicts during the flattening of the sysroot. In some cases binaries compiled for the wrong architecture ended up in the sysroot.
Further, a small logical error sneaked in, that lead to no packages being actually added to the sysroot at all.

## Removing gawk

The dependency on gawk is problematic since gawk itself requires a C++ toolchain. Using rules_distroless to create a sysroot for a C++ toolchain would therefore create a cyclic dependency.

I started with just writing tests for the existing logic to ensure that the replacements would behave similarly.
Then I rewrote the scripts to use builtin shell logic instead.

## Sysroot generation

toolchains_llvm expects a sysroot as a filegroup target. But it does not actually resolve the target but only uses the package of the target as the root of the sysroot. Files that are included in the filegroup must be present at repository fetch time.
This means that using the flatten rule to provide the sysroot is not possible!

Therefore we directly do the unpacking of the sysroot in the module extensions.
This in essence mirrors the behavior of flatten.

Since the unpacking happens independent on whether the sysroot is in the dependency graph of the built targets, it is by default disabled. Users can enable it by setting the `unpack` option to `True` in `apt.install`.
Unpacked sysroots are available for all architectures selected in `apt.sources_list`. The architectures are made available under distinc targets in the repo following the pattern `//unpack_<arch>` (see the example for more details).

Unpacking also considers mergedusr behavior. Again, we can not use the deb_postfix logic since it resolves at build time while we require the behavior at repository fetch time. To reduce duplication of the logic, I extracted the key parts from deb_postfix and reused them.

The final change in implementation was the scoping of mergedusr behavior. Instead of considering merged_usr in deb_import, I scoped it to apt.install.
This has the benefit, that if a user calls apt.install on a shared depset once with mergedusr and once without (for whatever reason) he will get the expected behavior.

The final two commits add examples and smoke testing to cover the basic behavior of the new API.

## Final remarks

I'm happy to discuss changes to any part of this PR. I'm not the absolute expert in Bazel. I tried some different ways to solve this and quite often hit my limits in terms of expertise while working on this. Long story short, there might be some awkward solutions to simple problems that I just did not know how to solve better. I'm happy to rework if somebody points them out to me!

If you want me to split the PR in smaller pieces (the above three sections for example) please drop me a note and I'll do the split.


LittleHuba:
Reworked the API to be more in line with the initial design in https://github.com/bazel-contrib/rules_distroless/issues/124

Though, I had to retain the unpacked sysroot in a subdirectory called "sysroot" to not get into issues with bazel and how the "sysroot" filegroup collects the files. But this is a minor difference to the initial design.

thesayyn:
This is something we are interested in supporting;  Would be nice if worked with existing apis instead of changing everything. 

See https://github.com/bazel-contrib/rules_distroless/pull/222

LittleHuba:
Thanks for the feedback @thesayyn.
I'll check out how I can reduce the changeset and will bring in pieces incrementally.
I see that the new API was merged in the meantime and some of my fixes were introduced by other PRs already.

It will take some time, but I'll come back with some new PRs to get this feature in!

PR #183, opened 2025-09-05
refactor: implement the new bzlmod api

thesayyn:
Fixes #123 #56 #66 https://github.com/bazel-contrib/rules_distroless/issues/124

PR #224, opened 2026-06-30
fix: Get CI green

blorente:
After #183 , some `bzl_library` targets were outdated. Fix them up.
Also, the file `distroless/toolchains.bzl` doesn't seem referenced by anything, and it looks like a leftover from the `WORKSPACE` days, so removed that as well.


loosebazooka (dismissed):

apt/private/BUILD.bazel
```diff
@@ -52,14 +52,17 @@ bzl_library(
...
 )
 
 bzl_library(
     name = "deb_import",
     srcs = ["deb_import.bzl"],
     visibility = ["//apt:__subpackages__"],
-    deps = ["@bazel_tools//tools/build_defs/repo:http.bzl"],
+    deps = [
+        ":pkgconfig",
+        "@bazel_tools//tools/build_defs/repo:http.bzl",
```
loosebazooka: it looks like http.bzl might not be necessary? #222 migrated off http.bzl in deb_import?

blorente: Oops, sorry about that! Thanks a lot, fixed in https://github.com/bazel-contrib/rules_distroless/pull/224/commits/e661f207333ea2eb894952790a4e2c0a4124ebf1

apt/private/BUILD.bazel
```diff
@@ -52,14 +52,16 @@ bzl_library(
...
-    deps = ["@aspect_bazel_lib//lib:strings"],
+    deps = ["@bazel_lib//lib:strings"],
 )
 
 bzl_library(
     name = "deb_import",
     srcs = ["deb_import.bzl"],
     visibility = ["//apt:__subpackages__"],
-    deps = ["@bazel_tools//tools/build_defs/repo:http.bzl"],
+    deps = [
```
loosebazooka: sorry I guess we should standardize style here too, single line this?

loosebazooka (approved):

loosebazooka (approved):

apt/private/BUILD.bazel
blorente: Happy to, done in https://github.com/bazel-contrib/rules_distroless/pull/224/commits/afa7525ebdcba05b05031d2c973891f8e168f76d.
I just run `bazel run //:buildifier.fix` before every push, I guess it doesn't have a strong opinion about this case (which I'm surprised by)

PR #227, opened 2026-07-01
[apt]: Drain all async download tokens before returning

Bojun-Feng:
#### Why I did it

`_resolve_downloads()` in `apt/extensions.bzl` starts multiple parallel downloads with `block=False`, then iterates through tokens calling `.wait()`. On the first successful download+decompress, it returns immediately without waiting on the remaining async tokens. Bazel 8.x requires all async work started by a module extension to be drained before the extension finishes, causing a hard error:

```
ERROR: Work pending after module extension @@rules_distroless+//apt:extensions.bzl%apt finished execution
ERROR: Pending asynchronous work after module extension finished execution
```

This affects every project using this branch with Bazel 8.x (e.g., the SONiC Bazel build).

Note this was previously opened as https://github.com/bazel-contrib/rules_distroless/pull/223 to the `new_api `branch, which got auto-closed since the `new_api `branch was merged into main and deleted.

#### How I did it

* Replace early `return` with deferred result pattern — store first successful result, continue loop to `.wait()` all remaining tokens, return after loop completes

#### How to verify it

1. Build any project that uses `rules_distroless@new_api` with Bazel 8.5.1 (e.g., `bazel build //...` from a repo using apt module extension)
2. Confirm no "Work pending" or "Pending asynchronous work" errors

#### Description for the changelog

Fix async download token drain in apt extension to comply with Bazel 8.x module extension requirements

### DIFF
--- a/apt/extensions.bzl
+++ b/apt/extensions.bzl
@@ -104,17 +104,23 @@ def _resolve_downloads(mctx, tokens, index_type, dist, comp, arch):
     Returns None for optional Contents when all attempts fail.
     """
     failed_attempts = []
+    result = None
     for (ext, cmd, url, url_idx, ext_name, output, token) in tokens:
         download = token.wait()
         decompress_r = None
+        if result != None:
+            continue
         if download.success:
             decompress_r = mctx.execute(cmd + [output])
             if decompress_r.return_code == 0:
                 target_triple = "{}/{}/{}".format(dist, comp, arch)
 
                 # Decompressed file lives in its own ext_name subdirectory
-                return ("{}/{}/{}/{}".format(target_triple, url_idx, ext_name, index_type), url, download.integrity, ext)
+                result = ("{}/{}/{}/{}".format(target_triple, url_idx, ext_name, index_type), url, download.integrity, ext)
+                continue
         failed_attempts.append((url + "/.../" + index_type + ext, download, decompress_r))
+    if result != None:
+        return result
 
     if index_type == "Contents":
         # Contents files are optional; some repositories (e.g. packages.cloud.google.com/apt)


thesayyn:
Doesn't this mean that now we have to wait for all other downloads to finish? possibly making the extension slower.

Bojun-Feng:
> Doesn't this mean that now we have to wait for all other downloads to finish? possibly making the extension slower.

Yes, waiting for all tokens does mean the extension blocks until the slowest one finishes. However, without this change, the extension hard fails on Bazel 8.x:
```
ERROR: Work pending after module extension @@rules_distroless+//apt:extensions.bzl%apt finished execution
ERROR: Pending asynchronous work after module extension finished execution
```
With the above error, the build doesn't complete at all. Bazel 8 requires all async tokens started by a module extension to be drained before the extension returns (See https://github.com/bazelbuild/bazel/pull/22883). The current early-return leaves tokens un-waited, which Bazel treats as a correctness violation and errors out.

Since `token.cancel()` isn't exposed in the Starlark API (only `.wait()` is available), I wasn't able to find an easy way to abandon in-flight tokens without triggering the error.

In practice, the performance penalty should be small: the tokens are downloading the same index file from the same server(s) in different compression formats, and failed formats are expected to return 404 quickly. And the `cached_format` mechanism already narrows subsequent runs to a single known-good extension, eliminating the drain overhead on repeat builds.

A potantial alternative is to implement a wave-based fallback (try one format at a time with all URLs in parallel, move to the next format only if the first wave fails). That would be a bigger change. Figured the minimal fix was worth proposing first since this currently breaks real builds (e.g. SONiC on Bazel 8.5.1).

thesayyn (approved):

PR #233, opened 2026-07-17
fix(e2e): Fix errors in `translate_dependency_set.bzl`

blorente:
There were two errors:
- The `//:flat` generated targets were never actually populated in the hub repository.
- In the test, we get `coreutils` from bullseye. However, bullseye ships `coreutils` with different versions for `arm64` and `amd64`, which caused hard-to-debug bugs.  We solve this by disallowing version differences, _unless_ they are different binary versions released from the same source (the `+bNN` suffixes).

Now, this doesn't on its own make the tests green. There is still another issue, where the `depends_on` carries both arm and amd versions of a dependency, and then `flatten` just squishes them all and the wrong one may win.

But that's a bigger fix, so I figured why not open this PR.


apt/private/translate_dependency_set.bzl
```diff
@@ -142,16 +143,43 @@ def _translate_dependency_set_impl(rctx):
...
             repo_name = util.sanitize(package_key)
             package = packages[package_key]
+            package_name = package["name"]
+
+            architectures_to_package_names[architecture].append(package_name)
 
+            source_version = version_lib.strip_binnmu(version)
+            recorded_version = package_names_to_versions.setdefault(package_name, source_version)
+            if recorded_version != source_version:
+                fail("""Two different source versions detected for package {name}: {v1} and {v2}.
```
thesayyn, line 170: Can this be relaxed later?

thesayyn (approved):

PR #239, opened 2026-07-29
fix: Handle linker scripts that appear as .so files

blorente:
Some `-dev` packages (e.g. libbsd-dev-0-11-7) have .so files which are actually linker scripts, possibly pointing to a different deb package. This PR is an attempt to solve that:

- In https://github.com/bazel-contrib/rules_distroless/commit/2270050d52792f145904a5c36d176829f3d0bbc3, we add a small parser for linker scripts. It's strict on purpose, as in it only allows certain keywords (`GROUP`, `INPUT`, `OUTPUT_FORMAT`, `AS_NEEDED`), but it can be expanded to handle more. It prints a list of files that need to be added as linker args.
- In https://github.com/bazel-contrib/rules_distroless/commit/2a2bd7a1df9e90e29d795ea5ddcb981cf447a1ee, we add the actual mechanism: We read all the .so files for the keywords `GROUP`, `INPUT`, `OUTPUT_FORMAT` etc, and parse them with the linker script parser. Then, we form `-Wl,--remap-inputs=...` cli flags by consulting the filemap. I'm very open to `--remap-inputs` not being the best way to model this. We have enough info so that a different solution is possible.


thesayyn (approved):
This is a nice cleanup!  Always wanted to do.

PR #242, opened 2026-07-30
Add tests for functions using gawk

LittleHuba:
Test the existing behavior before we replace gawk.
This helps with keeping behavior stable.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -22,9 +22,25 @@ jobs:
       - uses: actions/checkout@v4
       - uses: bazel-contrib/setup-bazel@0.15.0
       - run: bazel test //...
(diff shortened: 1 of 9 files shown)


LittleHuba:
I failed to reproduce this locally. So I'll extend the workflows to provide the error logs. Hopefully that gives some insights.

LittleHuba:
Fixed now by relaxing the tests to not consider ordering of entries in archives (see reasoning in code and commit messages).
I decided for this approach since from what I can see ordering of siblings within an archive does not actually matter in rules_distroless. Only order between files of different archives matters when doing the deduplication.

I'll leave the extension of the workflows in this PR, since this seems valuable for such cases where a local reproduction is not possible.

thesayyn (approved):

PR #244, opened 2026-08-03
Add mergedusr support to apt.install()

LittleHuba:
Introduces per-package/per-dependency-set mergedusr normalization for apt.install():

- apt.install() gains a `mergedusr` attribute. When set, the hub repo (translate_dependency_set) generates a mergedusr_compat symlink tar (bin/sbin/lib*/usr/... symlinks) using Bazel's native tar() rule with mtree directives, folded into the dependency set's :flat output.
- Packages are materialized per required mode (util.package_repo_name) so the same package can be built once with, and once without, mergedusr normalization, depending on which apt.install root(s) pull it in. compute_package_repo_modes() computes this scoping transitively so one apt.install(mergedusr=True) call cannot leak mergedusr behavior into unrelated apt.install calls that share dependencies.
- deb_import()'s existing mergedusr attribute is reused unchanged; only the naming/scoping around it is new.

### DIFF
--- a/apt/private/translate_dependency_set.bzl
+++ b/apt/private/translate_dependency_set.bzl
@@ -140,9 +140,9 @@ alias(
 # this scoping, a package's `depends_on` (which spans every architecture the
 # lockfile knows about) would leak foreign-architecture deps into a single
 # architecture's target, causing file duplication that confuses `flatten`.
-def package_deps_for_architecture(packages, package, architecture):
+def package_deps_for_architecture(packages, package, architecture, mergedusr = False):
     return [
-        "@" + util.sanitize(dep_key) + "//:data"
+        "@" + util.package_repo_name(dep_key, mergedusr = mergedusr) + "//:data"
         for dep_key in package["depends_on"]
         if packages[dep_key]["architecture"] in [architecture, "all"]
     ]
@@ -173,7 +173,7 @@ def _translate_dependency_set_impl(rctx):
 
         for (short_key, version) in dependency_set["sets"][architecture].items():
             package_key = short_key + "=" + version
-            repo_name = util.sanitize(package_key)
+            repo_name = util.package_repo_name(package_key, mergedusr = rctx.attr.mergedusr)
             package = packages[package_key]
             package_name = package["name"]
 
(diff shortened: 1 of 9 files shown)


apt/private/translate_dependency_set.bzl
```diff
@@ -81,16 +82,34 @@ filegroup(
...
         "{target_name}",
+        {flat_extra_tars}
     ],
     deduplicate = True,
     visibility = ["//visibility:public"],
 )
 """
 
+_MERGEDUSR_COMPAT_TMPL = """\
+tar(
```
thesayyn: can you explain why we need this?  shouldn't this come from the distro?

LittleHuba: Very good point. I somewhat blindly reimplemented the approach in deb_postfix:
https://github.com/bazel-contrib/rules_distroless/blob/3a63e6541784cbcbbaaedc38eddb85d9fa3dda1a/apt/private/deb_postfix.bzl#L36

I can't verify this for all different OS versions / distributions but at least for Ubuntu 24.04 LTS adding the "base-files" package leads to a working sysroot with correct mergedusr symlinks.

With this in mind, I guess we can close this PR as won't do.
But it raises the question to me whether we now have an inconsistency in rules_distroless? Or am I understanding the purpose of deb_postfix incorrectly?

thesayyn: A mergedusr attribute in apt.install makes sense and that should propagate all the way down to the target that export the deb files.

thesayyn: It’s just that I see extraneous code in here that the distros already fix

LittleHuba: Ahh okay understood. So the mergedusr flag should just fix package internals and users should use "base-files" to actually get the relevant symlinks.

I'll drop the symlink additions and leave the rest.

LittleHuba:
@thesayyn kindly requesting another review.
After this changeset my next PR would finally be the sysroot support.

LittleHuba:
@loosebazooka would you be willing to take over review of this PR in absence of @thesayyn?
I'd really like to progress on this PR to finish the sysroot support topic.

Background:
I'm slowly getting heat in my downstream project to remove the fork that patches the functionality in.

thesayyn:
Apologies; i was out PTO for two weeks.

thesayyn (approved):

loosebazooka:
whew :relieved:

PR #246, opened 2026-08-04
Fix and extend CI

LittleHuba:
Currently the CI only tests the main module.
This leads to broken examples and smoke tests.

This PR fixes the last broken example and extends the checks that are performed in the CI.

PR #248, opened 2026-08-04
fix: remove lockfile from smoke test

blorente:
Hopefully solving https://buildkite.com/bazel/bcr-presubmit/builds/38830/list?sid=019fc337-1d94-488b-bb73-e6e5a603210b&tab=output

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -20,8 +20,9 @@ jobs:
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: bazel-contrib/setup-bazel@0.15.0
+      - uses: bazel-contrib/setup-bazel@0.19.0
       - run: bazel test //...
+      - run: (cd e2e/smoke && bazel test //...)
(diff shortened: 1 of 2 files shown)


thesayyn (approved):

loosebazooka:
It seemed this was solved by locking the file in #246 ?

blorente:
@loosebazooka probably, yeah. Apologies, I didn't see that PR.

LittleHuba:
This PR tears a coverage gap into the smoke tests. Now the lock API is fully untested for Bazel 7.x
Instead of rebasing my PR #246 I kindly request to revert this PR and instead go ahead with #246.

blorente:
@LittleHuba You're absolutely right. I was hoping we could fix forward, but you're right that reverting is the right choice here: https://github.com/bazel-contrib/rules_distroless/pull/249

Apologies for the careless PR. I (wrongly) assumed that because we have migrated to the facts API, we wouldn't need this coverage anymore, and failed to consider Bazel 7.

PR #250, opened 2026-08-10
Revert "fix: remove lockfile from smoke test (#248)"

loosebazooka:
Reverts bazel-contrib/rules_distroless#248

Lets revert this and fix in #246


LittleHuba:
@loosebazooka can we merge #249 and #246? #249 does the same as this PR, so you would not require @thesayyn's approval.

PR #254, opened 2026-08-26
Sysroot support

LittleHuba:
This is finally the PR to add sysroot support to rules_distroless.

This PR consits of two commits:
1. The actual implementation for sysroot support
2. A refactoring of translate_dependency_set.bzl to align the tests that are performed on the different call paths

I think the second piece is crucial to not end up with bugs appearing on one path while they were already fixed on the other.

An example and smoke test were added to exercise the new functionality in CI and show users how to create the sysroot and inject it into a toolchain.


LittleHuba:
@thesayyn, @loosebazooka is it worth rebasing this PR to resolve the conflicts?

I'm willing to put the effort in to get this feature merged. But given that I did not hear any feedback for multiple weeks when this PR was ready, I'm unsure if this is still wanted from your side.

I just don't want to spend more time on this topic when this PR will never be merged in the end.

loosebazooka:
sorry, lemme take a look as it is now before you do any more work..

loosebazooka:
I was wondering if it would make sense to be its own rule in its own ruleset (to prevent @thesayyn and I from having to maintain this). 

But I'm leaning towards just including it here, it's small, very tightly coupled with rules_distroless and seems to be a significant improvement to prevent runtime breakages for our users. @thesayyn if you have no objections, lets get this rebased and I can do some reviews on it.

thesayyn:
rules_distroless already supports turning debian packages into importable cc_library targets with full transitive closure and understanding most debian packages. so i don't want to introduce an alternate  "please extract all the packages in the world into a directory" mode. 

We designed the sysroot support in a way so that every target declares what they depend on instead of the toolchain bringing every dependency that workspace depends on.

This is how we use rules_distroless for a custom gcc toolchain.

https://github.com/thesayyn/sonic-build-infra/tree/master/toolchains

loosebazooka:
okay that makes sense, lets close this for now, perhaps some documentation could be helpful here. But the functionality seems to exist.

LittleHuba:
Thanks for the feedback. That unfortunately means https://github.com/bazel-contrib/toolchains_llvm will not be able to support a sysroot generated through rules_distroless (without significant changes to one of the two rulesets).

@loosebazooka thanks for your suggestion. We are already using a workaround today that is based on a separate ruleset.
The problem is, that toolchains_llvm expects the sysroot before the analysis phase of Bazel starts. So the sysroot must be fully populated in a repository rule! That makes the current API of rules_distroless unsuitable for us.
We found a hacky workaround: https://github.com/eclipse-score/tooling/tree/main/bazel/rules/sysroot_from_lock
The problem is that we need some bzlmod resolution hacks to make it work: https://github.com/eclipse-score/communication/blob/main/MODULE.bazel#L243-L257 (e.g. directly depending on `ubuntu24_04_sysroot_resolve` and working on internals of your ruleset,...)
Bazel rightfully complains about this hack (https://github.com/eclipse-score/communication/actions/runs/35837447150/job/107104194233#step:4:19). So this is not really a feasible approach longterm.

Summary:

I don't see a way to put this logic outside of rules_distroless. Should your decision stand, we will have to drop our hack and rules_distroless. Keeping the hack when it is no longer temporary is not an option. This also means we can no longer describe within our project what our sysroot requires. If we don't want to fork and maintain rule_distroless, the new and only way to get the sysroot into toolchains_llvm will be to host the sysroot binary blob somewhere for Bazel to download and unpack.

Please don't get me wrong. I'm not trying to blame you into another decision. I'm just making transparent what the effect of your decision is for me as user.

If you have other ideas for a solution or change your mind, I'm happy to continue the discussion!

Thanks for your time!

loosebazooka:
We don't want to do duplicate extraction of the files to disk and we don't want to deal with the nuances of sysroot setup (symlink, stripping non-sysroot files, os level quirks) in this repository. And importantly we don't want to increase the public API surface that we need to maintain.

Let see what we can do here to prevent toolchains_llvm from being incompatible. 
1. It does look like perhaps toolchains_llvm is in the middle of a transition to a rule-based API, and if sysroot handling eventually follows this path, so maybe the resolution can occur later, more in line with how the rules_cc style of sysroot resolution works. That would maybe be an ideal long term solution, but wont help with anything now. Worth bringing up there.

2. It does appear that #268 makes `@<dependency_set>//:lock.json` available to your custom code in eclipse-score, whicih means maybe you can get rid of the `<repo>_resolve` target. This should resolve those warnings and allow bazel mod tidy and other tools to work cleanly.

This doesn't solve the other problem of "who maintains this" or "how to reuse this", but it still feels like maybe this doesn't belong here.

The improvements to translate_dependency_set.bzl,etc still seem useful, but those could be in a separate smaller PR.

thesayyn:
> The problem is, that toolchains_llvm expects the sysroot before the analysis phase of Bazel starts

This is a big problem that i don't want to keep encouraging. They should switch to rules_cc style sysroots where the sysroot directory can be provided by an action instead of bare files inside a repository rule.

PR #255, opened 2026-08-28
feat(apt): add OpenPGP signature verification for apt repository indices

loosebazooka:
- Verify InRelease / Release.gpg signatures via gpgv or sqv and enforce Release SHA256 checksums on package indices.
- Now requiring either gpg_keys or explicit allow_unsigned = True.
- Auto-dearmor ASCII-armored .asc keyrings (requires sq or gpg)

---

**What breaks:** Any existing apt.sources_list in a user's MODULE.bazel that does not specify gpg_keys will now fail Bazel evaluation with a clear error message.

**How users migrate:** Users must do one of two things:
1. Supply keyring(s) (recommended):
```
apt.sources_list(
    gpg_keys = ["//keys:debian-archive-keyring.asc"],
    ...
)
```
1.. Explicitly opt out of verification:
```
apt.sources_list(
    allow_unsigned = True,
    ...
)
```


apt/extensions.bzl
```diff
@@ -23,7 +23,167 @@ def _get_auth(mctx, urls):
         netrc = read_user_netrc(mctx)
     return use_netrc(netrc, urls, {})
 
-def _start_downloads(mctx, urls, dist, comp, arch, integrity, index_type, cached_format = None):
+def _find_pgp_verifier(mctx):
+    """Detect available OpenPGP verifier (gpgv or sqv)."""
```
thesayyn: should the user be able to provide a pgp binary?

loosebazooka: hah I'm still working through this. but yeah perhaps.

e2e/smoke/MODULE.bazel
```diff
@@ -35,6 +36,7 @@ apt.sources_list(
         "arm64",
     ],
     components = ["main"],
+    gpg_key = "//keys:debian-archive-keyring.asc",
```
thesayyn: can there be more than one key?

loosebazooka: yeah I believe this can be a set of keys.

apt/extensions.bzl
loosebazooka: well I think it needs to be gpgv or sqv for now because we know how those cli expect inputs. I imagine allowing any arbitrary verification provider is a little more complicated here unless we template out the verification command call.

e2e/smoke/MODULE.bazel
loosebazooka: this is actually update to `gpg_keys = [...]` to allow a few different keyrings

apt/extensions.bzl
```diff
@@ -23,7 +23,179 @@ def _get_auth(mctx, urls):
         netrc = read_user_netrc(mctx)
     return use_netrc(netrc, urls, {})
 
-def _start_downloads(mctx, urls, dist, comp, arch, integrity, index_type, source_id, cached_format = None):
+def _find_pgp_verifier(mctx):
```
thesayyn: can the pgp code here go into its own file?

loosebazooka: oh yeah that would make sense.

loosebazooka:
kk moved all the pgp stuff into it's own file.

thesayyn (approved):

PR #258, opened 2026-09-01
Re-add removed parameter `package_template` to `apt.install`

tempoz:
Somewhere in the shuffle of the refactoring of `apt.install`, this attr seems to have been dropped. It is vitally important if you want to use the `dpkg_statusd` rule provided by `//apt:defs.bzl`, as can be seen in the `distroless` repo here: https://github.com/GoogleContainerTools/distroless/blob/2958d510e6e7dcaa1f546042cfca98930a60a2ce/private/repos/deb/deb.MODULE.bazel#L21

This PR just re-adds the `package_template` parameter and re-pipes it through to `translate_dependency_set`, following the example of the PR that re-adds `mergedusr` here: https://github.com/bazel-contrib/rules_distroless/pull/244

I have tested this change locally in my project that depends on `rules_distroless` and confirmed that it works as intended.


apt/extensions.bzl
```diff
@@ -365,6 +368,7 @@ def _distroless_extension(mctx):
                             "amd64",
                             install.suites,
                             install.mergedusr,
+                            install.package_template,
```
loosebazooka: I don't think we need to plumb the package_template through the resolution queue. It's not affecting anything in this loop.

tempoz: fine with me; fixed.

loosebazooka:
okay this shoudl be fixed by #263 , lemme know if any behavior is missing

PR #259, opened 2026-09-01
Don't ignore the `include_transitive` attr in the `apt.install` rule

tempoz:
`apt.install` started silently ignoring the `include_transitive` attr. This PR corrects that.


apt/extensions.bzl
```diff
@@ -459,6 +461,7 @@ def _distroless_extension(mctx):
                     ("=", dep["Version"]),
                     arch,
                     suites,
+                    install.include_transitive,
```
loosebazooka: seems like this should be `include_transitive`, not install.include_transitive

tempoz:
```suggestion
                    include_transitive,
```

loosebazooka:
Adding a testing would be nice to check behavior. Either a simple smoke test in e2e or a unit test in package_resolution_test.bzl (hopefully prevening any future regression).

